PR #14, opened 2014-07-20
client/as-util: add Fedora point-of-contact to html (v2)

ignatenkobrain:
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>


hughsie:
I think we can drop this one now.

PR #39, opened 2015-03-14
configure.ac: gcab is a hard requirement

aruiz:
If gcab is not installed configure.ac would pass silently. This fixes it so that it enforces the dependency at configure time.


aruiz:
That was fast :D Thanks!

PR #51, opened 2015-06-05
Avoid using PATH_MAX

pinotree:
`realpath()` in POSIX.1-2008 supports `NULL` for the second argument, returning a newly allocated buffer. glibc has been supporting this mode for years already.

Thus, switch to this mode and avoid `PATH_MAX`, which is optional in POSIX and not provided on Hurd.

### DIFF
--- a/client/as-builder.c
+++ b/client/as-builder.c
@@ -40,7 +40,10 @@
  * as_builder_search_path:
  **/
 static gboolean
-as_builder_search_path (GPtrArray *array, const gchar *path, GError **error)
+as_builder_search_path (GPtrArray *array,
+			const gchar *path,
+			GHashTable *filter,
+			GError **error)
 {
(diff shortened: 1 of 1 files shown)


pinotree:
ping

hughsie:
_cleanup_free_ isn't right on the returned buffer as this calls g_free() on the result, not free()

pinotree:
> cleanup_free isn't right on the returned buffer as this calls g_free() on the result, not free()

OK, I will introduce a proper cleanup handler for it, then. After that, you might want to check `free` vs `g_free` in `libappstream-builder/asb-context.c` (`realpath` is used there).

pinotree:
Meh, I screwed up the branch, sorry :-/ I will create a new MR...

pinotree:
New MR: #59

PR #56, opened 2015-07-26
builder: Implement support for Solus .eopkg format

ikeydoherty:
This would essentially be for first review. I've already run tests against our archive and it runs perfectly. I'm unsure if we need to provide other bits of data to `AsbPackage`, I'd need your advice there :] Our notion of epoch is determined by distribution version, which right now is currently always 1, but this will be expanded in future to be correctly determined by the code.

The licensing I'm also needing feedback on. In Solus we support multiple license fields, and we're currently doing lots of scan/rebuilds to ensure we obtain the correct licenses being used by a codebase. Would bar (`|`) or `AND` separation of these identifiers be sufficient? (We're also moving to SPDX identifiers)

As for OOM bails, I'm unsure of your policy, thus, first review. 

In theory the check for eopkg binary is unnecessary, it just ensures a capable host is being used (one that knows eopkgs.)

### DIFF
--- a/libappstream-builder/asb-package-eopkg.c
+++ b/libappstream-builder/asb-package-eopkg.c
@@ -39,8 +39,6 @@
 #include <errno.h>
 #include <string.h>
 
-G_DEFINE_TYPE (AsbPackageEopkg, asb_package_eopkg, ASB_TYPE_PACKAGE)
-
 /**
  * Storage for eopkg metadata
  */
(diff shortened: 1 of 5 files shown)


libappstream-builder/asb-package-eopkg.c
```diff
@@ -0,0 +1,533 @@
...
+}
+
+/**
+ * asb_package_eopkg_new:
+ *
+ * Creates a new EOPKG package.
+ *
+ * Returns: a package
+ *
+ * Since: 0.1.0
```
ikeydoherty: Copy paste got me nowhere here :] I'll need to fix this in a version2.

hughsie: Right, all 0.4.2 i suppose...

hughsie:
As for SPDX, just use the full SPDX expression format, e.g. "GPL-2.0+ AND LGPL-2.0+" -- all the tools should understand that.

ikeydoherty:
AND over bar, got it. Alright I'll get a v2 together, thanks for looking, and I'll further glibificate it :] (there was quite the delta to start, 8 space vs 8-width tab, c99 vs glib c90, etc)

Issue #57, opened 2015-08-06
Unclear meaning of "Not enough <p> content before <ul>"


akurtakov:
appstream-util validate was showing this after adding second/third p section and sentence. The real error was that there were not enough characters. It would help a lot in fixing such issues if it states clearly that there are not enough characters preferably even with the number of missing characters.

ignatenkobrain:
validate-relax?
-- 

-Igor Gnatenko

hughsie:
Sure, makes sense. Have you got any other suggestions? The string is in https://github.com/hughsie/appstream-glib/blob/master/libappstream-glib/as-app-validate.c if you'd like to have a go at a patch.

kalev:
The commit above would change the error from:

```
/usr/share/appdata/eclipse.appdata.xml: FAILED:
• url-not-found         : <screenshot> url not found [http://kdaniel.fedorapeople.org/eclipsescreenshots/IDE.png]
• url-not-found         : <screenshot> url not found [http://kdaniel.fedorapeople.org/eclipsescreenshots/HELP.png]
• style-invalid         : Not enough <p> content before <ul>
Validation of files failed
```

to:

```
/usr/share/appdata/eclipse.appdata.xml: FAILED:
• url-not-found         : <screenshot> url not found [http://kdaniel.fedorapeople.org/eclipsescreenshots/IDE.png]
• url-not-found         : <screenshot> url not found [http://kdaniel.fedorapeople.org/eclipsescreenshots/HELP.png]
• style-invalid         : Content before <ul> is too short [210], at least 300 characters required
• style-invalid         : Content before <ul> is too short [63], at least 300 characters required
Validation of files failed
```

Is it any better?

akurtakov:
Way way easier to find issue if this is the output.

PR #58, opened 2015-08-06
Show a better error message when there's not enough text before <ul>

kalev:
Fixes https://github.com/hughsie/appstream-glib/issues/57


hughsie:
Does this still pass make check? If so, +1 from me. Thanks.

kalev:
Yup, make check still passes.

PR #61, opened 2015-09-12
Don't reallocate malloc'd memory with g_malloc

kalev:
glib's g_malloc and friends now always use malloc internally, as of glib
2.46: https://bugzilla.gnome.org/show_bug.cgi?id=751592

This means we no longer have to reallocate glibc malloc'd buffers and
can expect that g_free works fine on them.

### DIFF
--- a/libappstream-builder/asb-self-test.c
+++ b/libappstream-builder/asb-self-test.c
@@ -25,8 +25,6 @@
 #include <stdlib.h>
 #include <locale.h>
 
-#include "as-cleanup.h"
-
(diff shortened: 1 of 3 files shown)


hughsie:
Is free() now equal to g_free() as well?

kalev:
Yes. g_free() is just a trivial wrapper around free() now:

```
void
g_free (gpointer mem)
{
  if (G_LIKELY (mem))
    free (mem);
  TRACE(GLIB_MEM_FREE((void*) mem));
}
```

PR #75, opened 2016-01-05
Install gettext ITS rules

ueno:
Recent gettext release has a feature to allow consumer projects to
supply their own string extraction rules for XML files, in ITS format.
Gettext itself ships the rule for AppData, but it would be better
maintained in the upstream project.

See the gettext documentation for details:
http://www.gnu.org/software/gettext/manual/html_node/Preparing-ITS-Rules.html#Preparing-ITS-Rules

### DIFF
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -20,6 +20,9 @@ man_MANS +=						\
 endif
 endif
 
+itsdatadir = $(datadir)/gettext/its
+dist_itsdata_DATA = appdata.its appdata.loc
+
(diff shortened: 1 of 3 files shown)


ueno:
Sure, added *.metainfo.xml.

hughsie:
Do you think these kind of files belong in the main rpm, or the -devel subpackage? Thanks.

ignatenkobrain:
In devel subpkg.

hughsie:
Thanks!

PR #80, opened 2016-01-21
Add appstream-compose

alexlarsson:
This is a simple app that takes a prefix (default /usr) and a set of
app names. It looks for appdata, desktop and icon files with that
basename in $prefix/share/[appdata|applications|icons] and generates
appstream xml files + icons in (by default) $prefix/share/app-info.


hughsie:
We have this now.

Issue #70, opened 2015-12-16
Does not read MediaBaseUrl field in DEP-11 files


ximion:
Hi!
The `MediaBaseUrl` field in DEP-11 files is used as a "prefix" for media URLs, i.e. URLs of screenshots or icons of the `remote` type.
See the specification: http://www.freedesktop.org/software/appstream/docs/sect-AppStream-DEP11.html#spec-dep11-general

It fills in the same role `Origin` already has, and allows for easily switching the server which serves the media, which is a desired feature at Debian for derivatives as well as it will become useful in a setup where multiple mirrors exist for media content.

With the current implementation, asglib prevents GNOME-Software from showing any screenshots at all on Debian.
reference files for testing: ftp://ftp.debian.org/debian/dists/sid/main/dep11/

Cheers,
    Matthias

hughsie:
So, the reasons I'm not particularly happy about MediaBaseUrl are the following:

 * I'm unclear what problem it's trying to solve. Using MediaBaseUrl for the Fedora XML metadata the file size is reduced by 2%, but simply gzipping it it reduces by a further 60% (which is what we do with the XML metadata) and as a consequence it actually loads faster than the uncompressed metadata. If the argument is that it makes it easier to re-write the URL, you're no better off then just rewriting each URL as you still have to load and resave the entire file anyway.
 * It's unclear how to implement this. Does MediaBaseUrl supplement remote screenshots without a prefix (which incidentally makes them filenames or paths, not URLs), or does it overwrite all the remote screenshots? If it's the former then it's not actually a way to set up a screenshot mirror. Derivatives can't just rsync the screenshots and change the MediaBaseUrl as they need to produce AppStream metadata themselves to avoid the screenshots and the references in the files changing.
 * It makes it hard to merge files, due to unclear semantics of MediaBaseUrl we just have to remove it and make the URLs full, which kinda negates the purpose of the tag in the first place.
 * It's not clear what "media" it's prefixing; do remote icons get checked too?
 * It's going to increase the RSS for any files that use MediaBaseUrl and decrease the parsing speed for both files that do and files that don't.
 * I don't see how it can be used in a multiple mirror setup as you can only specify one URL for the MediaBaseUrl; it's not even useful for round-robin functionality.

ximion:
[...]

PR #96, opened 2016-03-03
Implement DEP-11 MediaBaseUrl support

robert-ancell:
This is in the DEP-11 spec:
https://www.freedesktop.org/software/appstream/docs/sect-AppStream-DEP11.html#spec-dep11-general

See https://github.com/hughsie/appstream-glib/issues/70


robert-ancell:
This is in the DEP-11 spec:

robert-ancell:
#https://github.com/hughsie/appstream-glib/issues/70

robert-ancell:
(I don't have write access, so I can't merge this myself).

hughsie:
You do now :)

PR #100, opened 2016-04-29
yaml: Make icon parser spec compliant

### LINKED ISSUE #98: Please support icon width/height for DEP-11 cached icons
Hi!
The DEP-11 spec switched from cached icons being a single string:
```YAML
ID: org.example.Test
Icon:
  cached: test_test.png
  stock: test
```
to them being a list of dicts to make it possible to include a width/height:
```YAML
ID: org.example.Test
Icon:
  cached:
    - width: 64
      height: 64
      name: test_test.png
    - width: 128
      height: 128
      name: test_test.png
```
(from [libas tests](https://github.com/ximion/appstream/blob/master/tests/test-yamldata.c#L184))
See https://www.freedesktop.org/software/appstream/docs/sect-AppStream-DEP11.html#field-dep11-icon

This feature can be implemented in a backwards-compatible way, as has been done in AppStream/libappstream already. Unfortunately, keeping forward-compatibility wasn't possible with this, and the adoption rate of this change seems to be high (I've already seen this being used "in the wild").
So, having this supported in GNOME Software too would be important. It should also speed up the time you need to search for icons ;-)
Thanks!
    Matthias

ximion:
This should resolve #98 which became a pretty high-priority item now... It also, while we're at it, implements the full spec for DEP-11-YAML icons, you never know when we'll need it...

The code is basically copied from libappstream.

Btw, issue 100 for asglib! \o/

### DIFF
--- a/data/tests/usr/share/app-info/yaml/aequorea.yml
+++ b/data/tests/usr/share/app-info/yaml/aequorea.yml
@@ -9,7 +9,10 @@ Name:
   C: Iceweasel
 Package: iceweasel
 Icon:
-  cached: iceweasel.png
+  cached:
+      - name: iceweasel.png
+        width: 64
+        height: 64
 Keywords:
   C:
     - browser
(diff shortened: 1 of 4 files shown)


ximion:
Pushed :-)

Issue #101, opened 2016-05-02
as-builder: reproducible resutls (icons.tar.gz)


DimStar77:
icons.tar.gz could benefit from an 'ordered' tarball, allowing to verify if the old/new tarball are equal to a previously generated one (and thus allowing to dynamically not publish a new tarball if no changes are detected).

Especiall in openSUSE, where the AppStream metadata is dynamically generated on repo-creation, this can be beneficial (the tarball often is the same, but the resulting md5 can differ)

Using gzip and tar (manually), one would use these parameters to achieve the right result:

tar --sort=name
gzip --no-name

Is this something as-builder could use too?

Relevant downstream bug report: https://bugzilla.opensuse.org/show_bug.cgi?id=977885

PR #102, opened 2016-05-02
Make archive creation deterministic

StefanBruens:
This should address issue #101.

### DIFF
--- a/libappstream-builder/asb-utils.c
+++ b/libappstream-builder/asb-utils.c
@@ -532,6 +532,9 @@ asb_utils_write_archive_dir (const gchar *filename,
 	if (files->len == 0)
 		return TRUE;
 
+	/* sort by filename for deterministic results */
+	g_ptr_array_sort (files, g_strcmp0);
+
 	/* write tar file */
 	return asb_utils_write_archive (filename, directory, files, error);
 }
--- a/libappstream-builder/asb-utils.c
+++ b/libappstream-builder/asb-utils.c
@@ -441,8 +441,10 @@ asb_utils_write_archive (const gchar *filename,
 	struct stat st;
 
 	a = archive_write_new ();
-	if (g_str_has_suffix (filename, ".gz"))
+	if (g_str_has_suffix (filename, ".gz")) {
 		archive_write_add_filter_gzip (a);
+		archive_write_set_filter_option (a, "gz", "timestamp", NULL);
+	}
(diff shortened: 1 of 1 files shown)


libappstream-builder/asb-utils.c
```diff
@@ -441,8 +441,10 @@ asb_utils_write_archive (const gchar *filename,
 	struct stat st;
 
 	a = archive_write_new ();
-	if (g_str_has_suffix (filename, ".gz"))
+	if (g_str_has_suffix (filename, ".gz")) {
 		archive_write_add_filter_gzip (a);
+                archive_write_set_filter_option (a, "gz", "timestamp", NULL);
```
ignatenkobrain: style: please use tabs

libappstream-builder/asb-utils.c
```diff
@@ -532,6 +534,9 @@ asb_utils_write_archive_dir (const gchar *filename,
 	if (files->len == 0)
 		return TRUE;
 
+        /* sort by filename for deterministic results */
```
ignatenkobrain: style: please use tabs (for 2 next lines as well)

libappstream-builder/asb-utils.c
```diff
@@ -441,8 +441,10 @@ asb_utils_write_archive (const gchar *filename,
 	struct stat st;
 
 	a = archive_write_new ();
-	if (g_str_has_suffix (filename, ".gz"))
+	if (g_str_has_suffix (filename, ".gz")) {
 		archive_write_add_filter_gzip (a);
+                archive_write_set_filter_option (a, "gz", "timestamp", NULL);
+	}
 	if (g_str_has_suffix (filename, ".bz2"))
 		archive_write_add_filter_bzip2 (a);
```
ignatenkobrain: what about bzip/xz/etc.? Don't we need to do the same for those?

StefanBruens: At least in libarchive the bz2 and xz filter have no timestamp option. *if* these formats have a timestamp(but I think they don't), these filters had to be enhanced in libarchive first.

Setting these options anyway would create an warning in libarchive.

libappstream-builder/asb-utils.c
StefanBruens: fixed

ignatenkobrain:
@StefanBruens thanks, merged!

PR #105, opened 2016-05-06
add missing "elementary" to list of project groups

decathorpe:
Most official elementary software sets the \<project_group\> tag content to "elementary", which is not a recognised project group. This renders nearly all their .appdata.xml files invalid when trying to validate them with appstream-util.

This change would include "elementary" in the list of recognised project groups (I guessed this has to be added to the  table at libappstream-glib/as-app.c:4750). The "homepage" URL used is "http://elementary.io/".

I assumed the list is intentionally alphabetically sorted, so I inserted the missing line according to that.

### DIFF
--- a/libappstream-glib/as-app.c
+++ b/libappstream-glib/as-app.c
@@ -4749,6 +4749,7 @@ as_app_parse_appdata_guess_project_group (AsApp *app)
 		const gchar *project_group;
 		const gchar *url_glob;
 	} table[] = {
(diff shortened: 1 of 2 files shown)


hughsie:
Whist valid, I think you also need to add the id to libappstream-glib/as-environment-ids.txt as well.

decathorpe:
Oops, I missed that one.

hughsie:
I've rebased and pushed this. Thanks!

PR #109, opened 2016-05-26
Fix g_ptr_array_sort use

kalev:
Looks like the current code used g_ptr_array_sort() incorrectly; fix untested. Noticed because gcc complained:

asb-utils.c: In function ‘asb_utils_write_archive_dir’:
asb-utils.c:538:27: warning: passing argument 2 of ‘g_ptr_array_sort’ from incompatible pointer type [-Wincompatible-pointer-types]
  g_ptr_array_sort (files, g_strcmp0);
                           ^~~~~~~~~
In file included from /usr/include/glib-2.0/glib.h:31:0,
                 from /usr/include/glib-2.0/glib/gprintf.h:21,
                 from /usr/include/glib-2.0/glib/gstdio.h:23,
                 from asb-utils.c:30:
/usr/include/glib-2.0/glib/garray.h:175:12: note: expected ‘GCompareFunc {aka int (*)(const void *, const void *)}’ but argument is of type ‘int (*)(const char *, const char *)’
 void       g_ptr_array_sort               (GPtrArray        *array,
            ^~~~~~~~~~~~~~~~

### DIFF
--- a/libappstream-builder/asb-utils.c
+++ b/libappstream-builder/asb-utils.c
@@ -508,6 +508,12 @@ asb_utils_add_files_recursive (GPtrArray *files,
 	return TRUE;
 }
 
+static gint
+my_pstrcmp (const gchar **a, const gchar **b)
+{
+	return g_strcmp0 (*a, *b);
+}
+
 /**
  * asb_utils_write_archive_dir:
  * @filename: archive filename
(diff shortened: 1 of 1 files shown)


kalev:
@StefanBruens Does this look right?

ignatenkobrain:
Isn't it just enough to add`(GCompareFunc)` for g_strcmp0?

kalev:
No, I don't think so: the pointers passed in to the comparison function have a layer of indirection and we need to dereference them before passing them to strcmp.

StefanBruens:
The fix is essentially correct, although I think it is nicer to do the casting inside the wrapper function. Taken from "man qsort":
```c
       static int
       cmpstringp(const void *p1, const void *p2)
       {
           /* The actual arguments to this function are "pointers to
              pointers to char", but strcmp(3) arguments are "pointers
              to char", hence the following cast plus dereference */

           return strcmp(* (char * const *) p1, * (char * const *) p2);
       }
```

PR #111, opened 2016-06-07
Add ASB Extractor Plugin for Sugar Activities

samdroid-apps:
Sugar Activities are a unique type of application desinged for Sugar.
Originally, they did not run outside of the Sugar DE.  However, they
now do and we will being shipping .desktop files for them [1].

We don't ship app data files.  It is much eaiser to just write a plugin
for ASB than write an app data file for every activity.  Having a
ASB plugin is also beneficial for us as it allows us to tag clearly
all Sugar Activities as being Sugar Activities.  This would allow us
to use App Stream as a basis for a Sugar Activity store and updater.

Additioanlly, it allows GNOME Software to have richer metadata when
accessing the Sugar Activities.  Sugar Activities already have
summaries, translations and icons, which are beneficial for
GNOME Software.

[1]  https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/321

### DIFF
--- a/libappstream-builder/plugins/asb-plugin-sugar-activity.c
+++ b/libappstream-builder/plugins/asb-plugin-sugar-activity.c
@@ -201,7 +201,6 @@ asb_plugin_process_actinfo (AsbPlugin *plugin,
 	tmp = g_key_file_get_string (actinfo, "Activity", "license", NULL);
 	if (tmp != NULL) {
 		as_app_set_project_license (app, tmp);
-		as_app_set_metadata_license (app, tmp);
 	}
 
 	tmp = g_key_file_get_string (actinfo, "Activity", "icon", NULL);
--- a/libappstream-builder/plugins/asb-plugin-sugar-activity.c
+++ b/libappstream-builder/plugins/asb-plugin-sugar-activity.c
@@ -88,13 +88,16 @@ asb_plugin_check_filename (AsbPlugin *plugin, const gchar *filename)
 
 static gboolean
 _asb_plugin_load_sugar_icon (const gchar *path,
-			     AsIcon *ic) {
+			     AsIcon *ic,
+			     AsApp *app,
+			     AsbPlugin *plugin) {
 	gsize len;
-	gchar *data = NULL;
-	gchar *tmp = NULL;
+	gboolean is_hidpi;
+	gchar *data, *name;
 	GBytes *bytes = NULL;
 	GString *string = NULL;
 	GRegex *regex = NULL;
+	GdkPixbufLoader *pbl = NULL;
 
 	if (!g_file_get_contents (path, &data, &len, NULL))
 		return FALSE;
@@ -102,25 +105,46 @@ _asb_plugin_load_sugar_icon (const gchar *path,
 	// Sugar icons have 2 XML entities, stroke_color and fill_color
 	regex = g_regex_new ("<!ENTITY stroke_color \"[^\"]+\">",
 			     0, 0, NULL);
-	tmp = g_regex_replace_literal (regex, data, len, 0,
-				       "<!ENTITY stroke_color \"#282828\">",
-				       0, NULL);
+	data = g_regex_replace_literal (regex, data, len, 0,
+					"<!ENTITY stroke_color \"#282828\">",
+					0, NULL);
 	g_free (regex);
 
 	regex = g_regex_new ("<!ENTITY fill_color \"[^\"]+\">",
 			     0, 0, NULL);
-	tmp = g_regex_replace_literal (regex, tmp, -1, 0,
-				       "<!ENTITY fill_color \"#FFFFFF\">",
-				       0, NULL);
+	data = g_regex_replace_literal (regex, data, -1, 0,
+					"<!ENTITY fill_color \"#FFFFFF\">",
+					0, NULL);
 	g_free (regex);
 
-	string = g_string_new (tmp);
+	// The PBL takes bytes/length, not null terminated string
+	string = g_string_new (data);
 	bytes = g_string_free_to_bytes (string);
 
-	as_icon_set_kind (ic, AS_ICON_KIND_EMBEDDED);
-	as_icon_set_data (ic, bytes);
-	as_icon_set_width (ic, 75);
-	as_icon_set_height (ic, 75);
+	is_hidpi = asb_context_get_flag (plugin->ctx, ASB_CONTEXT_FLAG_HIDPI_ICONS);
+	for (gint factor = 1; factor <= (is_hidpi ? 2 : 1); factor++) {
+		pbl = gdk_pixbuf_loader_new_with_type ("svg", NULL);
+		gdk_pixbuf_loader_set_size (pbl, 64 * factor, 64 * factor);
+		gdk_pixbuf_loader_write_bytes (pbl, bytes, NULL);
+		if (!gdk_pixbuf_loader_close (pbl, NULL))
+			return FALSE;
+
+		as_icon_set_kind (ic, AS_ICON_KIND_CACHED);
+		as_icon_set_prefix (ic, as_app_get_icon_path (AS_APP (app)));
+		as_icon_set_pixbuf (ic, gdk_pixbuf_loader_get_pixbuf (pbl));
+		as_icon_set_width (ic, 64 * factor);
+		as_icon_set_height (ic, 64 * factor);
+		if (is_hidpi) {
+			name = g_strdup_printf ("%ix%i/%s.png",
+			    64 * factor, 64 * factor,
+			    as_app_get_id_filename (AS_APP (app)));
+		} else {
+			name = g_strdup_printf ("%s.png",
+			    as_app_get_id_filename (AS_APP (app)));
+		}
+		as_icon_set_name (ic, name);
+	}
+
 	return TRUE;
 }
 
@@ -198,7 +222,8 @@ asb_plugin_process_actinfo (AsbPlugin *plugin,
 
 		if (g_file_test (path, G_FILE_TEST_IS_REGULAR)) {
 			ic = as_icon_new ();
-			if (_asb_plugin_load_sugar_icon (path, ic))
+			if (_asb_plugin_load_sugar_icon (
+				    path, ic, AS_APP (app), plugin))
 				as_app_add_icon (app, ic);
 		}
 	}
--- a/libappstream-builder/plugins/asb-plugin-sugar-activity.c
+++ b/libappstream-builder/plugins/asb-plugin-sugar-activity.c
@@ -91,14 +91,33 @@ _asb_plugin_load_sugar_icon (const gchar *path,
 			     AsIcon *ic) {
 	gsize len;
 	gchar *data = NULL;
+	gchar *tmp = NULL;
 	GBytes *bytes = NULL;
+	GString *string = NULL;
+	GRegex *regex = NULL;
 
 	if (!g_file_get_contents (path, &data, &len, NULL))
 		return FALSE;
-	printf ("ICON (got contents) = %s\n", path);
+
+	// Sugar icons have 2 XML entities, stroke_color and fill_color
+	regex = g_regex_new ("<!ENTITY stroke_color \"[^\"]+\">",
+			     0, 0, NULL);
+	tmp = g_regex_replace_literal (regex, data, len, 0,
+				       "<!ENTITY stroke_color \"#282828\">",
+				       0, NULL);
+	g_free (regex);
+
+	regex = g_regex_new ("<!ENTITY fill_color \"[^\"]+\">",
+			     0, 0, NULL);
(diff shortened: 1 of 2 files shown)


libappstream-builder/plugins/asb-plugin-sugar-activity.c
```diff
@@ -0,0 +1,286 @@
...
+static gboolean
+_asb_plugin_load_sugar_icon (const gchar *path,
+			     AsIcon *ic) {
+	gsize len;
+	gchar *data = NULL;
+	GBytes *bytes = NULL;
+
+	if (!g_file_get_contents (path, &data, &len, NULL))
+		return FALSE;
+	printf ("ICON (got contents) = %s\n", path);
```
ignatenkobrain: I guess something like g_debug() should be better

libappstream-builder/plugins/asb-plugin-sugar-activity.c
```diff
@@ -0,0 +1,286 @@
...
+
+	as_icon_set_kind (ic, AS_ICON_KIND_EMBEDDED);
+	bytes = g_bytes_new_take (data, len);
+	as_icon_set_data (ic, bytes);
+	as_icon_set_width (ic, 75);
```
hughsie: Is a 75x75 icon going to look good in gnome-software? Is it even loaded at that size?

libappstream-builder/plugins/asb-plugin-sugar-activity.c
```diff
@@ -0,0 +1,286 @@
...
+					 AS_MARKUP_CONVERT_FORMAT_SIMPLE,
+					 error);
+		if (desc != NULL)
+			as_app_set_description (app, NULL, desc);
+	}
+
+	tmp = g_key_file_get_string (actinfo, "Activity", "license", NULL);
+	if (tmp != NULL) {
+		as_app_set_project_license (app, tmp);
+		as_app_set_metadata_license (app, tmp);
```
hughsie: This isn't true; the metadata license is what the metadata is available under, and has to be a content license like CC0 rather than a code license like GPL-3.0+

libappstream-builder/plugins/asb-plugin-sugar-activity.c
```diff
@@ -0,0 +1,286 @@
...
+		// Icon might be "icon.svg" or might already have .svg suffix
+		path = g_build_filename (directory, tmp, NULL);
+		if (!g_file_test (path, G_FILE_TEST_IS_REGULAR)) {
+			const gchar *with_dotsvg;
+			with_dotsvg = g_strdup_printf ("%s.svg", tmp);
+			path = g_build_filename (directory, with_dotsvg, NULL);
+		}
+
+		if (g_file_test (path, G_FILE_TEST_IS_REGULAR)) {
+		    printf ("ICON (is file) = %s\n", path);
```
hughsie: g_debug again

libappstream-builder/plugins/asb-plugin-sugar-activity.c
```diff
@@ -0,0 +1,286 @@
...
+					     asb_package_get_basename (pkg));
+				return NULL;
+			}
+			asb_plugin_add_app (&apps, AS_APP (app));
+		} else if (_asb_plugin_is_linfo (filelist[i])) {
+			ret = asb_plugin_process_linfo (
+			    plugin, pkg, filename_tmp, AS_APP (app), error);
+
+			if (!ret)
+			    printf ("Bad activity.linfo: %s\n", filelist[i]);
```
hughsie: g_debug pls

libappstream-builder/plugins/asb-plugin-sugar-activity.c
```diff
@@ -0,0 +1,286 @@
...
+		as_app_set_metadata_license (app, tmp);
+	}
+
+	tmp = g_key_file_get_string (actinfo, "Activity", "icon", NULL);
+	if (tmp != NULL) {
+		g_autoptr(AsIcon) ic = NULL;
+		const gchar *directory;
+		const gchar *path;
+
+		printf ("ICON = %s\n", tmp);
```
hughsie: g_debug

libappstream-builder/plugins/asb-plugin-sugar-activity.c
```diff
@@ -0,0 +1,286 @@
...
+					G_KEY_FILE_NONE, error))
+		return FALSE;
+
+	if (!g_key_file_has_group (actinfo, "Activity"))
+		return FALSE;
+	if (!g_key_file_has_key (actinfo, "Activity", "bundle_id", error))
+		return FALSE;
+
+	as_app_add_category (AS_APP (app), "Sugar");
+	as_app_add_category (AS_APP (app), "SugarActivity");
```
hughsie: Are these Desktop Specification categories?

libappstream-builder/plugins/asb-plugin-sugar-activity.c
samdroid-apps: The icons are SVG, and are displayed in sugar as either 55x55 or 75x75.  Should I set the width and height differently to indicate svg?  They look like this in GNOME software:

screenshot from 2016-06-08 07-12-24

screenshot from 2016-06-08 07-15-15


Being sugar icons, they are also all recolourable.  I will look into recolouring them in this adapter to ensure that they are all the same black and white colors of the browse icons.  I can recolour them to a different combination if you want.

libappstream-builder/plugins/asb-plugin-sugar-activity.c
samdroid-apps: Sugar's activity.info files are more like .desktop files.  Doesn't that mean that they are part of the app, and the metadata has the same license.

Is a possible solution to just not indicate the metadata licence.

libappstream-builder/plugins/asb-plugin-sugar-activity.c
samdroid-apps: No.  Is there another way that we can mark sugar activities as being sugar activities in the appstream data?

libappstream-builder/plugins/asb-plugin-sugar-activity.c
samdroid-apps: I will clarify, that was GNOME software grabbing the svgs via the .destop file  I will test to ensure that the svgs work when embeded in the appdata.

samdroid-apps: Actually, the EMBEDDED svg data causes errors for GNOME Software :(

Is there a way to fix that?  I don't want to convert the SVGs to rasters, because it would benefit us in Sugar to recolour them (as we like to do to our icons).

libappstream-builder/plugins/asb-plugin-sugar-activity.c
hughsie: Not really; although we could argue that small snippets of text are licence and copyright-less, the long description really needs some licencing information. This was one of the reasons we needed an explicit metadata license for AppData files.

libappstream-builder/plugins/asb-plugin-sugar-activity.c
hughsie: So the AppStream spec states the icons get shipped as 64x64 PNGs, and additionally 128x128 if the --hidpi command like switch is used. We don't support loading SVGs from AppStream in gnome-software just due to performance reasons; showing 200 moderately complex svgs takes well over 200ms (which is our target from search to presentation). I do appreciate the Sugar ones are very simple. I don't think recoloring at runtime would be easily achievable.

samdroid-apps: Can applications ship multiple icons?  Could we ship a 64x64 pixbuf version along with a scalable version?

hughsie: I don't *think* so, that's probably a question for @ximion

samdroid-apps: I'm sure that we can talk about that later.  I just fixed up the thing to render the icons to a pixbuf.  Commit bf7ef656b242d9ab272b99fd7f660de927d6c182

libappstream-builder/plugins/asb-plugin-sugar-activity.c
samdroid-apps: Ok, I fixed it up to not set the metadata license.  Commit ec204bc7191f079b28c85551e0200c61aa450a0d

libappstream-builder/plugins/asb-plugin-sugar-activity.c
ximion: An application can not ship multiple different icons and it is not allowed to ship an SVG icon either, for the reasons @hughsie explained.
But you don't need that, actually, since appstream-builder will render the SVG to PNG icons (I suppose ^^).
I also don't see the use case for having apps have multiple different icons, if anything, it adds more confusion to the users. The only case where this is probably useful is for monochromatic low-res icons for accessibility reasons.

Btw, on a more general note, I don't think this plugin is great, since it is exclusive for appstream-builder, meaning only stuff using that tool to generate AppStream metadata will benefit from it. Additionally, converting formats usually means loss of information, since AppStream metainfo files usually can contain much more data than the source format that you convert into them. Having a "magic" conversion removes the need for writing metainfo files for the respective upstream project, making it ship less metadata in comparison.
It also prevents other tools which handle metainfo files from picking them up when a sugar app is installed.
So IMHO this is a nice workaround to show those apps quickly, but you should really write metainfo files for these apps.
Cheers,
    Matthias

samdroid-apps: Making metadata files for these activities seems hard - I'm not sure how we would get the translations working given that we don't install a gettext domain or anything like that.

Additionally, this icon code recolours the icons.  That is useful for GNOME software, because otherwise our icons look inconsistent.

But I really don't care that much about GNOME software.  It is already getting some metadata from the `.dekstop` files.

I am really making this patch for the benefit of Sugar.  Unfortunately, it is not optimal for Sugar.  64x64 png icons don't really fit our design language of 75x75 grid cells with recoloured svg icons.  Not using our metadata files means that we don't get our metadata.  I will look into an alternative solution.

samdroid-apps: Actually, @ximion, do you approve of the plugin for the gnome shell extension?  The one that is currently in the tree?

ximion: @samdroid-apps Haha, well:
From a metadata-purity point of view I disapprove that plugin's presence too. It's special-casing for a specific type of component and we really, really should not encourage that.

From a realistic perspective though, we have some metadata, and showing incomplete metadata is better than having no metadata. So I consider this plugin to be a good thing if the long-term plan is to get rid of it.
Now that I'm aware that this thing exists:
@hughsie what's the state of this feature? I see you added a "shell extension" component-type. Is this still an experimental feature? If not, it would make sense to add it too the spec, and I would also implement the metadata-extraction workaround in asgen, so Arch, Debian and Ubuntu can benefit from this.
That being said, "shell-extension" is a terrible name, because "shell" is waaay to generic. It's not a Plasma shell extension, not an extension to Bash or Zsh ;-)
Probably defining all of these things as component-type "addon" and have them extend the shell makes sense... (GS could special-case that).
Anyway, this isn't the best place to discuss this, I hope I'll think about moving it to the Mailinglist (otherwise -> IRC).

hughsie:
So, going back a bit; what's the aim of this PR? I thought it was to allow not-yet-installed sugar apps to appear in gnome-software?

samdroid-apps:
2 motivations really;

1.  Sugar is adding more support for running activities outside of sugar.  Therefore,  it would also be advantageous to have sugar activities display correctly in GNOME software.

2.  Sugar's currently uses the windows software distribution model.  That is, people go to activities.sugarlabs.org (an ancient AMO fork - looking yuck now), and download and run bundles off the internet.  Yuck.  Wouldn't it be better to just use appstream/packagekit and have it go through distro repositories.  We would want a little more metadata to have it best integrated with sugar (aka, svg icons, a "ModernCollaberation" flag for activities that support modern, no-telepathy-tubes collaboration, etc.).

Anyway, I can open a thread on the ML if you prefer.

hughsie:
Thinking about this some more, @ximion is right; I think we do need AppData files upstream for this to work correctly; there is just not enough metadata in the existing sugar files. I think we could work on the spec enhancements required for shipping scalable icons, but I think sugar apps shipped by the distribution really need the extra metadata too. I know this isn't a small amount of work.

samdroid-apps:
Awesome.  I'll start some mailing list threads about these issues.

Replaced by https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/331 on the sugar side (which just generates the appdata.xml files).

ximion:
Great, thank you!

PR #112, opened 2016-06-10
Metainfo preference changes

ximion:


hughsie:
We certainly want to fall back!

ximion:
I figured that ^^
I'll add it to my todo list, probably something to finish this evening.

data/appdata-xml.m4
```diff
@@ -36,7 +36,7 @@ AU_DEFUN([APPDATA_XML],
          AS_IF([test "x$enable_appdata_validate" = "xyes"],
                [AC_MSG_ERROR([AppData validation was requested but appstream-util was not found])])])
 
-  AC_SUBST([appdataxmldir], [${datadir}/appdata])
+  AC_SUBST([appdataxmldir], [${datadir}/metainfo])
```
hughsie: I don't think we're quite ready for this change in distros; at least in fedora i'd need to make some changes to the base packages.

libappstream-glib/as-store.c
```diff
@@ -2788,7 +2788,7 @@ as_store_search_per_user (AsStore *store,
 						 "flatpak",
 						 "exports",
 						 "share",
-						 "appdata",
+						 "metainfo",
```
hughsie: So this code path isn't going to be used for the next gnome-software release onwards, so i think we can ignore this for now.

libappstream-glib/as-store.c
```diff
@@ -2822,10 +2822,10 @@ as_store_search_per_user (AsStore *store,
 			return FALSE;
 	}
 
-	/* AppData */
+	/* AppData / MetaInfo */
 	if ((flags & AS_STORE_LOAD_FLAG_APPDATA) > 0) {
 		g_autofree gchar *dest = NULL;
-		dest = g_build_filename (g_get_user_data_dir (), "appdata", NULL);
+		dest = g_build_filename (g_get_user_data_dir (), "metainfo", NULL);
```
hughsie: We need to fall back here, although I can do this.

data/appdata-xml.m4
ximion: Oh, that valuable info! I can confirm that Arch, Debian, Ubuntu and OpenSUSE are ready for this change. I assumed Fedora was ready too, since appstream-builder supports this...

hughsie: Do you have a list of packages that ship files in metainfo handy? I can do the base-filesystem changes myself but i need to add some reasoning to the commit.

libappstream-glib/as-store.c
ximion: Jup (I only consider the first patch to be merge-ready... Originally I wanted to update the second patch to have a fallback, but somehow forgot about this over the weekend...)

data/appdata-xml.m4
ximion: We have a couple of them on Debian, and this is slowly making its way upstream too (I would have switched KDE already, but we need to be sure all distros support it before making this change).
Scribus in Debian and Ubuntu would be an example: https://packages.debian.org/experimental/scribus-ng
(I also know at least three other packages, but I am not sure how helpful Debian packages are for you)
In any case, it's safe to assume that this will make its way upstream rather quickly.

Conan-Kudo:
@hughsie @ximion If this is already done, shouldn't this PR be closed?

ximion:
It's only partially done, I kept it open so we don't forget about it.
(AFAIR the change to always install into metainfo dirs needs to happen, and adjusting the docs slightl, that's it. Also, I think it was blocked on RHEL not being up to date until recently).

jbicha:
Could we please update `appdataxmldir` now?

There is now a Debian warning with the [lintian checker](https://lintian.debian.org/tags/appstream-metadata-in-legacy-location.html) for the deprecated `/usr/share/appdata/` directory.

ximion:
This patch is a bit harsh by today's standards, back then I thought we could at least for Flatpak (since it was young) start with only one supported directory.
Now, that Flatpak is more widely used with the current configuration, we clearly shouldn't do that.

Switching `appdataxmldir` should be very painless though, and not cause any issues (the metainfo dir is supported for a long time already)

hughsie:
I've pulled in the first chunk of this manually in https://github.com/hughsie/appstream-glib/commit/9df56f26f047e6f59be5265cf97eb54a41f6a94e. I think we're good now, thanks.

ximion:
Yes, looks good :-)
Thanks!

PR #114, opened 2016-06-30
Add an AS_STATE_PURCHASABLE constant so we can support purchasing apps

robert-ancell:
I haven't completed the other side of this patch in GNOME Software but I'm pretty sure this state is the way to go. If you're in agreement it is useful to land this now so G-S etc can depend on a released version that contains the state.


hughsie:
LGTM.

PR #115, opened 2016-07-05
Add an AS_STATE_PURCHASING constant

robert-ancell:


hughsie:
Isn't this just INSTALLING?

robert-ancell:
The state changes I was working on were:

PURCHASABLE -> PURCHASING -> AVAILABLE -> INSTALLING -> INSTALLED

In GNOME Software, I expect that a purchase dialog is shown during the PURCHASING state. I added this state because it's like installing in that it could take a long time (might have to go to some bank backend (sure, in theory it would be fast, but banks).

We could save on states by doing:

PURCHASABLE -> AVAILABLE -> INSTALLING -> INSTALLED

I guess it depends on what the states are for. For GNOME Software, both are workable. Without PURCHASING another part of the code would not know that a purchase transaction is in progress, which may or may not be important!? In GNOME Software the only case I can think of is the "Buy" wouldn't change once you clicked on it (to "Buying").

hughsie:
It does seem a bit odd to make the user go through the two-step mental model of "click buy", "click install" -- I mean; if they purchased the app they probably wanted to install it. The fact they had to purchase the app is just a detail in the installation progress. I'd say these states are valid:

PURCHASABLE -> INSTALLING -> INSTALLED (payment successful, install successful)
PURCHASABLE -> INSTALLING -> PURCHASABLE (payment did not succeed)
PURCHASABLE -> INSTALLING -> UNKNOWN -> AVAILABLE (payment succeeded, install not so much -> angry user)

robert-ancell:
In terms of the UI, I had G-S automatically do the install once it got to the AVAILABLE state so from the users point of view there's not two steps.

hughsie:
I guess if it makes things easier; I suppose the UI code is easier this way too.

PR #117, opened 2016-07-09
AsApp: add API to control noshow project groups

cosimoc:
These map to the NotShowIn fields of desktop files, and can be used from
clients such as gnome-software to make decisions based on such
information.

### DIFF
--- a/libappstream-glib/as-app-desktop.c
+++ b/libappstream-glib/as-app-desktop.c
@@ -270,6 +270,16 @@ as_app_parse_file_key (AsApp *app,
 		if (tmp != NULL && tmp[0] != '\0')
 			as_app_add_pkgname (app, tmp);
 
+	/* NotShowIn */
+	} else if (g_strcmp0 (key, G_KEY_FILE_DESKTOP_KEY_NOT_SHOW_IN) == 0) {
+		list = g_key_file_get_string_list (kf,
+						   G_KEY_FILE_DESKTOP_GROUP,
+						   key,
+						   NULL, NULL);
+
+		for (i = 0; list[i] != NULL; i++)
+			as_app_add_noshow_project_group (app, list[i]);
+
 	/* OnlyShowIn */
 	} else if (g_strcmp0 (key, G_KEY_FILE_DESKTOP_KEY_ONLY_SHOW_IN) == 0) {
 		/* if an app has only one entry, it's that desktop */
--- a/libappstream-glib/as-app.c
+++ b/libappstream-glib/as-app.c
@@ -83,6 +83,7 @@ typedef struct
 	GPtrArray	*bundles;			/* of AsBundle */
 	GPtrArray	*translations;			/* of AsTranslation */
 	GPtrArray	*vetos;				/* of string */
+	GPtrArray	*noshow_project_groups;		/* of string */
 	AsAppSourceKind	 source_kind;
 	AsAppState	 state;
 	AsAppTrustFlags	 trust_flags;
@@ -395,6 +396,7 @@ as_app_finalize (GObject *object)
 	g_ptr_array_unref (priv->bundles);
 	g_ptr_array_unref (priv->translations);
 	g_ptr_array_unref (priv->vetos);
+	g_ptr_array_unref (priv->noshow_project_groups);
 
 	G_OBJECT_CLASS (as_app_parent_class)->finalize (object);
 }
@@ -423,6 +425,7 @@ as_app_init (AsApp *app)
 	priv->bundles = g_ptr_array_new_with_free_func ((GDestroyNotify) g_object_unref);
 	priv->translations = g_ptr_array_new_with_free_func ((GDestroyNotify) g_object_unref);
 	priv->vetos = g_ptr_array_new_with_free_func (g_free);
+	priv->noshow_project_groups = g_ptr_array_new_with_free_func (g_free);
 
 	priv->comments = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
 	priv->developer_names = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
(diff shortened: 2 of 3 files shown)


hughsie:
@ximion are you interested in supporting this in the XML specification too?

ximion:
@hughsie So far we did not support this "artificially hide stuff on desktop A, but show it on desktop B" stuff on purpose.
Reason is that stuff that is that desktop-specific should not be shown in a software center at all, and should probably be tagged with a non-GUI-app component type (if it is that desktop-specific).
By allowing something like `NotShowIn`, we also effectively prohibit users from uninstalling stuff that they might want to remove on the system.
Furthermore, the semantics of this tag aren't really clear, e.g. when using the CLI client, we want to show all metadata, so naming stuff `NotShowIn` is really misleading (this could probably be fixed by having a differently-named tag).
Also, esp. for `NotShowIn` the thing breaks as soon as another desktop appears - `OnlyShowIn` is a bit better there...

So, in summary: I would like to have a solid use-case for supporting this (so some kind of bug where adding support for this is the only option to fix the issue), especially since so far we deliberately did not add such a feature.

mcatanzaro:
The use case is we have Screen Reader (orca) and GNOME Color Manager appearing under the Installed tab in Software, but neither of these are applications in GNOME.

 * For the case of Orca, it's a system service in GNOME rather than an application. It doesn't show any desktop file, and you start it by flipping a switch in System Settings or from the a11y shell menu. But it really is an application that requires a desktop file in other desktops, as otherwise there'd be no way to start it.
 * For the case of GNOME Color Manager, it's a component of System Settings (gnome-control-center) in GNOME rather than an application, but it makes sense that it should be possible to run it in other desktops as a standalone application.

Note that uninstalling GNOME Color Manager uninstalls all of GNOME and drops you to a virtual terminal (as I discovered to my dismay a few weeks ago), so this is really a critical issue.

But I agree that we need to think about the implications of having apps that appear or disappear from Software depending on which desktop you are running. We probably don't want users switching from GNOME to KDE, uninstall GNOME Color Manager, and then not being able to get back into GNOME, for instance.

cosimoc:
@ximion See https://bugzilla.gnome.org/show_bug.cgi?id=768601 for the gnome-software side of this. Note that I did not implement any policy at all in how appstream-glib makes use of this data -- it just exposes it to the upper layers so that clients such as gnome-software can have logic that makes a decision based on the data.

matthiasclasen:
I've always considered NotShowIn/OnlyShowIn to be really terrible misfeatures of the desktop entry spec, for all the reasons ximion mentions.

Cases like "I'm orca, I can't decide if I'm an app or not" don't justify bringing this misfeature back, imo. Orca needs to be split into a service component that is an integral part of the desktop, and whatever is app-like in it (I don't see it) can live separately. Covering such application design issues up in the desktop file is a recipe for pain, as we've learned over many years.

At the end of the day, users have a surprisingly high tolerance to clutter (just look at peoples gnome2 desktops, with all the screen edges plastered with icons), and a much lower tolerance for hard-to-explain system behavior ('why is this app showing up when I run kde, but not in gnome. I know it is installed!')

ximion:
@mcatanzaro I agree with @matthiasclasen that the apps like Orca should not be shown if they are really a desktop-specific module. You could declare them to be an `addon` to `gnome-systemsettings` in AppStream, which would be accurate.
For the case when the application is runnable from other desktops, we should show it as separate app in software centers.

The thing is: I want AppStream to be a common way to assign metadata to any software component and retrieve it in a generic way. If I add logic to AppStream which sometimes shows apps depending on some conditions, this would work contrary to the project's goals.
The maximum thing I could think of adding is some kind of `designed_for_desktop` tag, but even for that I would be quite sceptical.
(For .desktop files, the logic does make some sense, for a metadata format with the aim to be descriptive and not control how something is (not) displayed or works, it doesn't)

That being said, since your patch doesn't touch AppStream itself at all, it would be okay with me if @hughsie added it - but it also wouldn't solve your problem at all. It would make the application disappear as soon as it is installed, but still have it shown before it is installed. Which is IMHO the worst possible way to handle things.

As for "make the backend powerful and have the frontends decide what of the functionality they want to use": I am all for that idea. But there are limits to it. For example, if the backend exposes functionality which encourages bad behavior at the frontend side, then it would be a really bad idea to add such functionality in the first place. That's the reason why @hughsie rejected adding functionality to export log output of package-managers via a PackageKit interface for ages, and subsequently, no (GUI) frontend ran into the trap of showing console output in its UI.

So, what to make out of it? I really don't want the `OnlyShowIn`/`NotShowIn` logic in AppStream, because I can only see this causing a lot more problems in future. But I also don't want to categorically reject it - if after some time we see that not having it creates really big problems which can not be solved by other means, I would obviously reconsider this (since that would show that not adding this feature in the past had been a mistake).
I am sorry I can't give a more positive reply to this - let's wait for time to prove the (non) necessity of this.

mcatanzaro:
I think we just need to split the gnome-color-manager/orca desktop files into separate packages... or remove them entirely, as they're not needed for GNOME and these are GNOME packages. I got permission from Joanie to delete orca's desktop file. @hughsie, what do you think we should do with color manager?

hughsie:
@mcatanzaro I've already split up the gnome-color-manager upstream .spec file example (a new -utils subdirectory) -- I'll push that into F25 at some point this week.

mcatanzaro:
@hughsie I think you need three subpackages; we shouldn't have multiple desktop files in the same package, since that means installing one app would unexpectedly pull in two others, right?

mcatanzaro:
@ximion I think we're going to drop this issue and fix apps instead. Does that sound OK, @cosimoc?

@hughsie is already handling gnome-color-manager. See https://bugzilla.gnome.org/show_bug.cgi?id=735761 for orca.

cosimoc:
@mcatanzaro Fine by me.

PR #124, opened 2016-08-02
Handle monitor events for individual files added with as_store_from_file()

smcv:
If you use `as_store_from_file()` to load individual `.xml.gz` files into a store, and replace those files with `g_file_set_contents()` or an equivalent atomic-overwrite from another process, then the file change notification emits

    As-WARNING **: no path data for /.../xmls

and doesn't actually emit the `::changed` signal. This appears to be because the code in `AsStore` that watches the monitor assumes it will be monitoring the `.xml.gz` file's parent directory, not the file itself.

While digging into this I noticed there is also some redundant monitoring going on: `as_store_load_app_info_file()` results in adding a monitor for the `.xml.gz` file itself, even though in that code path we are already going to monitor its parent directory. The file monitoring is only needed for `as_store_from_file()`.

### DIFF
--- a/libappstream-glib/as-store.c
+++ b/libappstream-glib/as-store.c
@@ -108,6 +108,7 @@ static gboolean	as_store_from_file_internal (AsStore *store,
 					     GFile *file,
 					     AsAppScope scope,
 					     const gchar *arch,
+					     AsStoreWatchFlags watch_flags,
 					     GCancellable *cancellable,
 					     GError **error);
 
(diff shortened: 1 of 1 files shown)


hughsie:
I assume make check still passes with this? If so, LGTM, thanks.

smcv:
0.5.x version also available: https://github.com/smcv/appstream-glib/tree/05x-monitor

smcv:
> I assume make check still passes with this?

It does when I add a patch to make the DEP-11 tests conditional on having enabled DEP-11, which I'll send as a separate PR.

smcv:
I see you've added me as a contributor. Should I rebase onto `no-dep11` (#125) and push?

Do you want the 0.5.x version too? (That's what I initially developed, this one is a rebase with conflict fixes for s/id_prefix/scope/ - I can send a separate PR if you'd like to review the 0.5.x code separately)

hughsie:
If you want 0.5 i'm happy with you just backporting that without a PR. I'm kinda easy going. Just don't break continuous or ebassi will hunt you down.

PR #125, opened 2016-08-02
Skip DEP-11 tests when DEP-11 is not enabled

smcv:


hughsie:
LGTM, please push to master.

PR #130, opened 2016-08-18
as-app-validate: Check if icon_name is NULL

justinjoy:
When 'as_icon_get_{file}name' return NULL, 'as_app_validate_icons' function
will show Critical Warning by calling 'g_str_has_prefix'. But it wouldn't be
a critical and sometimes, it could be NULL.

### DIFF
--- a/libappstream-glib/as-app-validate.c
+++ b/libappstream-glib/as-app-validate.c
@@ -709,7 +709,8 @@ as_app_validate_icons (AsApp *app, AsAppValidateHelper *helper)
 		break;
 	case AS_ICON_KIND_LOCAL:
 		icon_name = as_icon_get_filename (icon);
-		if (!g_str_has_prefix (icon_name, "/")) {
+		if (icon_name == NULL ||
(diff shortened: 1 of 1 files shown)


libappstream-glib/as-app-validate.c
```diff
@@ -709,7 +709,8 @@ as_app_validate_icons (AsApp *app, AsAppValidateHelper *helper)
 		break;
 	case AS_ICON_KIND_LOCAL:
 		icon_name = as_icon_get_filename (icon);
-		if (!g_str_has_prefix (icon_name, "/")) {
+		if (icon_name == NULL &&
```
ignatenkobrain: shouldnt it be `!=`?

justinjoy: Oops, I was confused.

I want to check if icon_name is `NULL` because `NULL` is also invalid so `==` is right, but shouldn't be `&&` with `g_str_has_prefix`.

Thank your for reminding me.

PR #133, opened 2016-08-22
trivial: Use correct autoptr cleanup on yaml error when loading from a file

superm1:
It seems this was needed in a second place.  This fixes calling from appstream-util validate.

> $ appstream-util validate ./yaml.yml
> ./yaml.yml: scanner error: mapping values are not allowed in this context at ln:2 col:5


superm1:
I forgot to mention in the pull request, can you also backport that to 0_5_X?

hughsie:
Done.

PR #134, opened 2016-08-26
Wip/hughsie/meson

jpakkane:
Some cleanups to Meson configuration.

You might also consider moving more preprocessor defines from command line arguments to `config.h` or equivalent as much as possible. This is beneficial in many ways such as:

- less places for state to exist
- when everything is in config.h it is easy to inspect
- smaller command line argument lists are easier to inspect and are faster to process (not usually noticeable in small projects but good practice anyway)
- passing command line arguments with shell metacharacters (i.e. double quotes) through various backends and their invocations is *hard* and there may be interesting bugs, especially on less often used backends

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -3,12 +3,13 @@
 #
 # Licensed under the GNU Lesser General Public License Version 2.1
 
-project('appstream-glib', 'c')
+project('appstream-glib', 'c', version : '0.6.2')
 
-as_major_version = '0'
-as_minor_version = '6'
-as_micro_version = '2'
-as_version = '@0@.@1@.@2@'.format(as_major_version, as_minor_version, as_micro_version)
+as_version = meson.project_version()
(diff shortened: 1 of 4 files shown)


meson.build
```diff
@@ -3,11 +3,11 @@
...
-
-as_major_version = '0'
-as_minor_version = '6'
-as_micro_version = '2'
+project('appstream-glib', 'c', version : '0.6.2')
+varr = meson.project_version().split('.')
+as_major_version = varr[0]
+as_minor_version = varr[1]
+as_micro_version = varr[2]
 as_version = '@0@.@1@.@2@'.format(as_major_version, as_minor_version, as_micro_version)
```
ignatenkobrain: this also should be converted then

ignatenkobrain: basically it's same as meson.project_version()

PR #135, opened 2016-09-05
Add -lang package as a dependency in order to properly detect transla…

DimStar77:
…tions

In openSUSE, the -lang package is 'only' recommended, so that users CAN save the
space if they don't need them (and using bundles there is also a method to only install
one language instead of all supported ones).

Unfortunately, with this technique, gnome-software keeps on pretending that no application
is localized in openSUSE.

### DIFF
--- a/libappstream-builder/asb-package-rpm.c
+++ b/libappstream-builder/asb-package-rpm.c
@@ -333,6 +333,10 @@ asb_package_rpm_ensure_deps (AsbPackage *pkg, GError **error)
 			*tmp = '\0';
 		asb_package_add_dep (pkg, dep_no_qual);
 	}
+        /* Add the corresponding -lang package as a dependency */
+        tmp = g_strconcat (asb_package_get_name (pkg), "-lang", NULL);
+        asb_package_add_dep (pkg, tmp);
+        g_free (tmp);
 out:
 	rpmtdFreeData (td);
 	rpmtdFree (td);


libappstream-builder/asb-package-rpm.c
```diff
@@ -333,6 +333,8 @@ asb_package_rpm_ensure_deps (AsbPackage *pkg, GError **error)
 			*tmp = '\0';
 		asb_package_add_dep (pkg, dep_no_qual);
 	}
+        /* Add the corresponding -lang package as a dependency */
+        asb_package_add_dep (pkg, g_strconcat(asb_package_get_name(pkg), "-lang", NULL));
```
ignatenkobrain: this will not work on Fedora.

DimStar77:
from my tests it also won't hurt though - as the dep does not really have to exist

libappstream-builder/asb-package-rpm.c
hughsie: It also leaks the string, and needs a space before both brackets.

hughsie:
Not apposed to this, I agree it's pretty harmless on other distros.

hughsie:
Would have preferred a g_autofree, but this looks okay to me. +1, please push to master.

DimStar77:
just squashed the commits properly; I decided to use tmp as it was already an existing variable that at this moment was 'free to be used'

DimStar77:
@hughsie I don't have merge access in this repo - if this is fine, It'd be nice of you to merge it then

PR #136, opened 2016-10-03
Make as_app_kind_to_string output current names of enumerated values

robert-ancell:
The values were changed in ea8b160c8f05ae018dcf54181204cc5740c2aa1c

### DIFF
--- a/libappstream-glib/as-app.c
+++ b/libappstream-glib/as-app.c
@@ -151,7 +151,7 @@ const gchar *
 as_app_kind_to_string (AsAppKind kind)
 {
 	if (kind == AS_APP_KIND_DESKTOP)
-		return "desktop";
+		return "desktop-application";
 	if (kind == AS_APP_KIND_CODEC)
 		return "codec";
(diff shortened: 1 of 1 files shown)


hughsie:
I thought about this; if we do this then we'll create metadata that older versions of appstream-glib can't process. I think we can merge this, bu maybe not just yet.

ximion:
If this gets merged, we can also switch the default install path for metainfo files to `/usr/share/metainfo`, since that path has been supported for a really long time now (much longer than the component names).
But whatever works best for your schedule is fine :-) - there is no rush to update.

robert-ancell:
I'll leave this PR open so people know about the reason for the current code. Feel free to merge when it's appropriate to do so.

ximion:
It's important to note that while the AppStream reference implementation has a fallback for `desktop` <-> `desktop-application`, this does not happen for `webapp`, because unlike `desktop`, `webapp` was never a recognized component type name in AppStream.
So, without this change, appstream-glib might generate invalid AppStream metadata.

PR #141, opened 2016-12-09
Add Geary to the app id fallbacks

nielsdg:
Geary recently renamed its app id to follow flatpak conventions. Adding a fallback to appstream.

For more info, see [bug 766196](https://bugzilla.gnome.org/show_bug.cgi?id=766196) on GNOME bugzilla.


nielsdg:
Thanks!

Op vr 9 dec. 2016 om 10:39 schreef Richard Hughes <notifications@github.com>

PR #143, opened 2016-12-29
Add support for "+" operator at the end of SPDX license identifiers

kalev:
SPDX v2.0 and later support an unary "+" operator suffix at the end of
license identifiers, e.g. "CC-BY-SA-3.0+". This commit adds support for
tokenizing and detokenizing this and adds new tests.


hughsie:
Does this need any similar fix (for the tokenization->URLifiying) in gnome-software?

kalev:
Yes. I have a gnome-software fix ready to go if the appstream-glib side makes sense to you.

hughsie:
It does make sense; please merge this :)

kalev:
Thanks!

PR #149, opened 2017-01-26
as_store_remove_app_by_id: Don't use-after-free

smcv:
We can't call methods on the app after we can no longer be sure that
we hold a reference to it.

---

This caused a crash in testing some code based on appstream-glib, in a test that exercises what happens when an AppStream `.xml.gz` file changes. In this particular project not everything in the AppStream is immediately useful (we're looking at a subset), so we don't hold a separate ref on AsApp objects that aren't of interest - I think that's why I saw this but gnome-software wouldn't.

### DIFF
--- a/libappstream-glib/as-store.c
+++ b/libappstream-glib/as-store.c
@@ -1217,7 +1217,7 @@ as_store_add_app (AsStore *store, AsApp *app)
 	g_ptr_array_add (priv->array, g_object_ref (app));
 	g_hash_table_insert (priv->hash_unique_id,
 			     (gpointer) as_app_get_unique_id (app),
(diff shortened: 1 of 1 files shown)


smcv:
An alternative way to do this would be to make sure that `hash_unique_id` holds its own ref to its value (with a free-function), instead of using a "borrowed" ref from the `array`. The other `hash_foo` hashes do all seem to hold refs.

hughsie:
I think holding the ref in all cases is probably safer...

smcv:
Fine, I'll respin this to be in those terms.

smcv:
Redone as requested.

PR #151, opened 2017-01-28
Add a destructor for AsRefString

kalev:
AsRefString uses a singleton hash table for deduplicating strings. This
is all "possibly lost" memory according to valgrind as the hash table
never gets destroyed.

This commit adds a custom destructor so that we can clean up the
singleton hash table as well and avoid unnecessary noise in valgrind
output.

### DIFF
--- a/libappstream-glib/as-ref-string.c
+++ b/libappstream-glib/as-ref-string.c
@@ -64,6 +64,12 @@ as_ref_string_get_hash_safe (void)
 	return as_ref_string_hash;
 }
 
+static void __attribute__ ((destructor))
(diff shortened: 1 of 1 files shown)


hughsie:
Looks good, but does this work on all compilers?

kalev:
__attribute__ ((destructor)) should be supported on gcc and clang at least. There's a private gconstructor.h header file in glib we could copy to get additional msvc and sunpro compiler support too.

hughsie:
Okay, lets wait for the person using msvc to provide the patch :)

PR #152, opened 2017-01-29
Fix the build with gperf 3.1

kalev:


kalev:
An alternative way to fix this would be to add a configure check to detect the function signature, similar to how is done in https://github.com/systemd/systemd/commit/c9f7b4d356a453a01aa77a6bb74ca7ef49732c08

hughsie:
Could we do the 3rd section fix like https://github.com/libguestfs/libguestfs/commit/004de6cf45cfe66ca166cbdcb1c3125a2a81eb31 ?

Richard.

kalev:
You mean merge as-tag.c and as-tag.gperf together into one file?

hughsie:
Well, if we do that it's a bit more tricky to support with-and-without gperf. Ideally not all of as-tag.c, but just _as_tag_from_gperf()

kalev:
Pushed a new approach that follows what systemd is doing after our irc discussion.

PR #160, opened 2017-03-16
preserve 0x prefix in the version string less than 0xff

timchen119:
For hex version less than `0xff`, the `0x` prefix was being removed accidentally and later it may converted to wrong value. A use case is that in https://bugs.launchpad.net/bugs/1651552 when fwupdmgr parse firmware.inf in the .cab file with version `0x1a`, it will first convert to `1a` and later `1a` is being treat as non-numeric content. The pull request should fix this issue by preserving the `0x` prefix in the version string less than 0xff.

### DIFF
--- a/libappstream-glib/as-self-test.c
+++ b/libappstream-glib/as-self-test.c
@@ -4006,11 +4006,17 @@ as_test_utils_version_func (void)
 		AsVersionParseFlag flags;
 	} version_from_uint32[] = {
 		{ 0x0,		"0.0.0.0",	AS_VERSION_PARSE_FLAG_NONE },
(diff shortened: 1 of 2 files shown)


hughsie:
Can you include some tests in as-self-test.c please.

timchen119:
Added some tests, thanks!

hughsie:
I've pushed 1a327ebc8e45b07be0f193f768848f1d1be06433 which fixes things in a slightly different way. Can you verify this fixes the original bug you see please. Thanks!

timchen119:
Confirmed 1a327eb fixes the original bug. Thanks!

Issue #127, opened 2016-08-11
Validation fails for component console-application


chronitis:
I tried validating (`appstream-util validate-relax --nonet org.test.consoleapp.metainfo.xml`) a console application metainfo file as per <https://www.freedesktop.org/software/appstream/docs/sect-Metadata-ConsoleApplication.html>

The metainfo file contained

```
<?xml version="1.0" encoding="UTF-8"?>
<component type="console-application">
    <id>org.test.consoleapp</id>
    <metadata_license>MIT</metadata_license>
    <project_license>MIT</project_license>
    <name>ConsoleApp</name>
    <summary>A test application</summary>
    <provides>
        <binary>consoleapp</binary>
    </provides>
</component>
```

This failed with

```
org.test.consoleapp.metainfo.xml: FAILED:
• attribute-invalid     : <id> has invalid type attribute
• markup-invalid        : <id> does not have correct extension for kind
Validation of files failed
```

It appears `as_app_kind_to_string` and friends do not handle this component type and it is treated as unknown.

hughsie:
I think this is fixed in git, thanks @ximion.

PR #161, opened 2017-03-17
as-app-validate: Allow MIT for metadata_license

jbicha:
According to the [specification](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-metadata_license), MIT is acceptable for metadata_license.

### DIFF
--- a/libappstream-glib/as-app-validate.c
+++ b/libappstream-glib/as-app-validate.c
@@ -1044,6 +1044,8 @@ as_app_validate_is_content_license_id (const gchar *license_id)
 		return TRUE;
 	if (g_strcmp0 (license_id, "@FSFAP") == 0)
 		return TRUE;
+	if (g_strcmp0 (license_id, "@MIT") == 0)
+		return TRUE;
 	if (g_strcmp0 (license_id, "&") == 0)
 		return TRUE;
 	if (g_strcmp0 (license_id, "|") == 0)


hughsie:
Urgh, @ximion was there any discussion on this? I need to run this past the RH legal team.

ximion:
The MIT license was requested incredibly early, I think it has been on the list since the inception of the `metadata_license` tag. Weird that nobody noticed so far, in https://github.com/hughsie/appstream-glib/issues/127 apparently there even was no warning for it.

Anyway, I didn't even know that this is something that needs discussion, since the spec pretty much states "all free permissive licenses" and just gives examples for what that means. So in theory, people can put any license in there, in practice libappstream(-glib) limit the choices to a list of predefined "OK" licenses.
I had no idea that appstream-glib didn't accept MIT... I think quite a few projects are using that.

hughsie:
The issue isn't so much that it's a permissive license, it needs to be a valid "content" license rather than a code license.

ximion:
At time the spec says "all permissive" - should we restrict that to "only licenses from this list"?
Btw, I think MIT/Expat shouldn't cause problems, as it adds no restrictions which make mixing the metadata with other data problematic.

hughsie:
I think it certainly has to be from a hardcoded list. What license is the DEP-11 file, for instance? The Fedora AppStream file is listed as `CC0 OR CC-BY`... etc

ximion:
The DEP-11 stuff is handled like the rest of the archive metadata: "For the individual licenses refer to the package files (debian/copyright)".

Conan-Kudo:
I'm not sure it makes sense to have "code" licenses allowed for content, even ones like the MIT/Expat license.

@ximion The behavior for DEP-11 sounds like a cop-out. Aren't they supposed to render the license information in the AppStream metadata?

ximion:
@Conan-Kudo No, the spec actually only defines the `metadata_license` tag for metainfo files, not for the collection data.
Adding it wouldn't be a problem, but the gain of doing so wouldn't be huge (unless there is one of the permitted licenses which demands its name to be shown in the processed data, which I think isn't the case).
Also, the processed data contains information from other sources (.desktop files) as well, so I am not sure if we can just declare it FSFAP or whatever license is defined in the metainfo file.

Conan-Kudo:
@ximion As far as I know, CC-BY should be satisfactory for this case, and everyone actually knows what that is...

ximion:
@Conan-Kudo There is no "one license fits all" many projects have different requirements...
From a perspective of Debian, all CC-* licenses are a bit nasty because they demand a copy of the license to be shipped with the source, and many upstream projects (who might only use CC-0 for the metainfo file) don't comply with that. So we have to file bugs etc., and additionally copy the whole license text into debian/copyright for documentation.

Conan-Kudo:
@ximion What?

As far as I understand it, that's not true.

For example, [CC-BY](https://creativecommons.org/licenses/by/4.0/legalcode) states that a link to the resource is sufficient (see §3(a)(1) and §3(a)(2)).

That applies across the entirety of the CC family.

ximion:
> For example, CC-BY states that a link to the resource is sufficient (see §3(a)(1) and §3(a)(2)).

And that link is usually nowhere to be seen.

> The debian/copyright thing is a special burden from Debian. You could have just as easily added it to your common-licenses package

The licenses there are only extended in very rare events when the amount of packages using a particular license becomes high enough. This is not (yet?) the case for CC-BY licenses.

> And the GNU licenses require verbatim copies shipped with it, which Debian blatantly ignores...

That is just wrong, it's one of the things our ftpmasters and QA tools look out for. You won't find any GPL-licensed source package without full GPL copy, and no binary package without debian/copyright file linking to the full GPL copy in common-licenses, which is available on each and every Debian system and can not be removed (as it's part of base-files, which also defines the root directory layout).

Conan-Kudo:
@ximion In the case of app centers like GNOME Software, the link *is* present. In the case of Debian, they're just doing it wrong.

And if your ftp masters *are* looking out for this, then why do none of them ship the GPL license that comes with the source code? Instead, you point them to common-licenses. There are scores of software targeting Debian and derivatives that don't ship the license *because* they believe it is sufficient to just point to some other copy. For CC licenses, this is fine. For GNU licenses, it is not.

ximion:
> In the case of app centers like GNOME Software, the link is present.

Whether some other random tool shows a link has nothing to do with what the license text demands, and you know that.

> And if your ftp masters are looking out for this, then why do none of them ship the GPL license that comes with the source code?

Ehh, what? All sources ship the GPL license, if one is missing we ask upstream to include it.

> Instead, you point them to common-licenses.

No, that is only for binary packages. And even there it's legit, because there is no Debian system in existence that does not ship a copy of the GPL, simply because you can not remove it without breaking the system or making something that is non-Debian (i.e. by patching base-files).
That is sufficient. Think of common-licenses as a deduplication thing for texts in debian/copyright. The debian/copyright file does *not* replace the full copy of the license upstream has to ship with the code and that must be present in an orig.tar.* tarball.

So, what was the actual topic of this bug again?

hughsie:
I think close this. MIT isn't a content license.

jbicha:
Then we need to fix the specification.

https://codesearch.debian.net/search?q=metadata_license.*MIT

jbicha:
Unlike someone using say GPL for the metadata, Debian and Ubuntu does show apps (Telegnome, Beignet, etc.) using MIT in the GNOME Software app.

ximion:
The MIT license was recognized as valid content license in libappstream since basically forever, so changing that back would be quite invasive.

hughsie:
I'm just telling you what our legal team advised me. I'm not sure I can go into much more details here, but happy to expand in private.

PR #163, opened 2017-03-28
Copy hash table keys from AsApp to avoid them being used after they've been freed

robert-ancell:
The hash tables currently reference the keys in the AsApp object, however this
fails in two cases:

1. If you insert two apps with the same ID into the same table the second app
replaces the first but the key is from the original one (confirmed in glib
source code). The first app is unreffed leaving the table to contain an invalid
key (or may occur later if a reference is held elsewhere).

2. There are two tables that contain GPtrArrays of apps. If AsApp in the array
that was used as the key is unreffed then the hash table again contains a freed
key.

I'm assuming these keys weren't copied to save memory, if that is important then
the keys will have to be reference counted or atoms used.

### DIFF
--- a/libappstream-glib/as-store.c
+++ b/libappstream-glib/as-store.c
@@ -929,7 +929,7 @@ as_store_remove_app (AsStore *store, AsApp *app)
 	g_signal_emit (store, signals[SIGNAL_APP_REMOVED], 0, app);
 
 	/* only remove this specific unique app */
-	apps = g_hash_table_lookup (priv->hash_id, g_strdup (as_app_get_id (app)));
+	apps = g_hash_table_lookup (priv->hash_id, as_app_get_id (app));
 	if (apps != NULL) {
 		g_ptr_array_remove (apps, app);
 
--- a/libappstream-glib/as-store.c
+++ b/libappstream-glib/as-store.c
@@ -929,7 +929,7 @@ as_store_remove_app (AsStore *store, AsApp *app)
 	g_signal_emit (store, signals[SIGNAL_APP_REMOVED], 0, app);
 
 	/* only remove this specific unique app */
-	apps = g_hash_table_lookup (priv->hash_id, as_app_get_id (app));
+	apps = g_hash_table_lookup (priv->hash_id, g_strdup (as_app_get_id (app)));
 	if (apps != NULL) {
 		g_ptr_array_remove (apps, app);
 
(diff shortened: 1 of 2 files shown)


robert-ancell:
This was found by running the following through valgrind on Ubuntu Zesty (code copied from gnome-software):

    g_autoptr(AsStore) store = as_store_new ();
    g_autoptr(GError) error = NULL;

    if (!as_store_load (store,
                        AS_STORE_LOAD_FLAG_IGNORE_INVALID |
                        AS_STORE_LOAD_FLAG_APP_INFO_SYSTEM |
                        AS_STORE_LOAD_FLAG_APP_INFO_USER |
                        AS_STORE_LOAD_FLAG_APPDATA |
                        AS_STORE_LOAD_FLAG_DESKTOP |
                        AS_STORE_LOAD_FLAG_APP_INSTALL,
                        NULL,
                        &error)) {
         g_printerr ("Failed to load store: %s\n", error->message);
         return EXIT_FAILURE;
    }

The app in this case that was causing the issue is org.gnome.Software.desktop (which I found quite funny to be crashing itself).

robert-ancell:
This is the main crasher in Ubuntu with gnome-software in Ubuntu Zesty.

libappstream-glib/as-store.c
```diff
@@ -929,7 +929,7 @@ as_store_remove_app (AsStore *store, AsApp *app)
 	g_signal_emit (store, signals[SIGNAL_APP_REMOVED], 0, app);
 
 	/* only remove this specific unique app */
-	apps = g_hash_table_lookup (priv->hash_id, as_app_get_id (app));
+	apps = g_hash_table_lookup (priv->hash_id, g_strdup (as_app_get_id (app)));
```
hughsie: Dup with the lookup? This looks odd...

hughsie:
Can you also write a self test (that crashes) that's also fixed by this please. Something little and self contained in as-self-test -- thanks.

hughsie:
Just the self tests then we're good to roll up and push. Thanks.

robert-ancell:
Ack, still working on it. I think I can modify it to not copy the keys, but it's not playing nice :(

robert-ancell:
The test case is odd, because it is a quite specific set of actions. It also runs fine normally, unless you are checking the output of valgrind (or happen to have a system that crashes on the invalid memory access).

I gave up making the hash tables not copy the keys, it just makes too flaky.

robert-ancell:
The actual case on my system that was causing this:
- The store loading the gnome-software metadata from the .desktop file (priority -1)
- The store loading the .appdata.xml file for gnome-software, and merging it with the .desktop file content (priority -1)
- The store loading the gnome-software metadata from the YAML files, replacing the previous data (priority 0).

At least once in the above the merging and replacing referred to the new AsApp object using the ID key from one of the previous AsApp objects.

smcv:
> I gave up making the hash tables not copy the keys, it just makes too flaky.

If the keys are "borrowed" from a struct member of the values, `g_hash_table_replace()` is the usual trick. But if it isn't obvious how to do it correctly, copying the keys is much simpler - obviously-correct is usually better than correct-but-subtle.

PR #172, opened 2017-06-15
tests: Fix 'gcab' executable requirement

iainlane:
I noticed that we were trying to run this test even though we hadn't built the file. It depends on 'gcab' (exe) but we check for libgcab the library.

I guess you could do this dynamically too, to avoid the need for the extra define. Let me know if you want that instead.

### DIFF
--- a/data/tests/meson.build
+++ b/data/tests/meson.build
@@ -17,7 +17,9 @@ if get_option('enable-builder')
     ],
     command : [gcab, '--create', '--nopath', '@OUTPUT@', '@INPUT@'],
   )
+endif
 
+if conf.has('HAVE_GCAB_EXE')
   firmware202 = custom_target('firmware-2.0.2',
     output : 'colorhug-als-2.0.2.cab',
     input : [
@@ -26,5 +28,6 @@ if get_option('enable-builder')
       'firmware/2_0_2/firmware.metainfo.xml'
     ],
     command : [gcab, '--create', '--nopath', '@OUTPUT@', '@INPUT@'],
+    build_by_default: true,
   )
 endif
--- a/libappstream-glib/as-self-test.c
+++ b/libappstream-glib/as-self-test.c
@@ -3602,7 +3602,7 @@ as_test_store_merges_local_func (void)
 static void
 as_test_store_cab_func (void)
 {
-#ifdef HAVE_GCAB
+#if defined(HAVE_GCAB_EXE)
 	gboolean ret;
 	const gchar *src;
 	g_autoptr(GError) error = NULL;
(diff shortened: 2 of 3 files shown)


hughsie:
Is there a case where we want the library but not the binary, or vice versa? It seems to me we just want a big HAVE_GCAB that checks for the binary and the library.

iainlane:
We only actually use the executable in as-builder, and meson.build already checks for it in that case. libappstream-glib itself doesn't require it AFAICS - just its testsuite for this one test.

Can do it like that if you want, it's just asking for a little bit more than is strictly required.

(Actually I just realised that `HAVE_GCAB_EXE` implies `HAVE_GCAB`, so we don't need the `&&` in the current proposal.)

iainlane:
Sorry, I wasn't clear - do you want me to change it like you described or do you accept my argument? :-)

hughsie:
build_by_default seems to be upsetting Travis. Is there any way we can avoid bumping the required meson version?

ignatenkobrain:
Actually, 0.38.0 is quite old... I don't think there is problem with bumping that.

ximion:
@hughsie Bump it, there shouldn't be an issue with that for Debian (except for backporting, obviously, but I think Meson will receive a backport anyway).

Conan-Kudo:
@ximion Do you have a solution to make it pass on Debian Stretch?

```
$ docker run -t -v `pwd`:/build appstream-glib-debian-stretch ./contrib/ci/build_and_test.sh -Denable-rpm=false -Denable-dep11=false

WARNING: You are using 'ANSI_X3.4-1968' which is not a a Unicode-compatible locale.

WARNING: You might see errors if you use UTF-8 strings as filenames, as strings, or as file contents.

WARNING: Please switch to a UTF-8 locale for your platform.

The Meson build system

Version: 0.37.1

Source dir: /build

Build dir: /build/build

Build type: native build


Meson encountered an error in file meson.build, line 1, column 0:

Meson version is 0.37.1 but project requires >=0.38.0.



The command "docker run -t -v `pwd`:/build appstream-glib-debian-stretch ./contrib/ci/build_and_test.sh -Denable-rpm=false -Denable-dep11=false" exited with 1.
```

ximion:
@Conan-Kudo Yes. Install a newer Meson from stretch-backports :-P

hughsie:
@iainlane can you rebase this please. Thanks.

iainlane:
@hughsie is it right that since d301f0cc4e467bdd7d2988056820fafbe3c6cfbf we don't actually require gcab any more? If so, this can be closed.

hughsie:
My kind of bug. Thanks for verifying.

PR #176, opened 2017-07-07
store: Provide async variants of load functions

fmuellner:
The existing functions may block, which isn't ideal in IO-sensitive
contexts. To address this, add async variants that run the functions
in a thread.

### DIFF
--- a/libappstream-glib/as-store.c
+++ b/libappstream-glib/as-store.c
@@ -2920,6 +2920,70 @@ as_store_load_path (AsStore *store, const gchar *path,
 					path, cancellable, error);
 }
 
+static void
+store_load_path_thread (GTask *task,
+			gpointer source_object,
+			gpointer task_data,
+			GCancellable *cancellable)
+{
+	AsStore *store = source_object;
+	const char *path = task_data;
+	GError *error = NULL;
+	gboolean success;
+
+	success = as_store_load_path (store, path, cancellable, &error);
+	if (error)
+		g_task_return_error (task, error);
+	else
+		g_task_return_boolean (task, success);
+}
+
+
+/**
+ * as_store_load_path_async:
+ * @store: a #AsStore instance.
+ * @path: A path to load
+ * @cancellable: a #GCancellable.
+ * @callback: A #GAsyncReadyCallback
+ * @user_data: Data to pass to @callback
+ *
+ * Asynchronously loads the store from a specific path.
+ *
+ * Since: 0.7.2
+ **/
+void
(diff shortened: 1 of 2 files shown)


hughsie:
Can you use 8-char tabs please. Otherwise LGTM.

fmuellner:
Ping?

libappstream-glib/as-store.c
```diff
@@ -3157,6 +3221,70 @@ as_store_load (AsStore *store,
...
+		   gpointer task_data,
+		   GCancellable *cancellable)
+{
+	AsStore *store = source_object;
+	AsStoreLoadFlags flags = GPOINTER_TO_INT (task_data);
+	GError *error = NULL;
+	gboolean success;
+
+	success = as_store_load (store, flags, cancellable, &error);
+	if (error)
```
hughsie: `!= NULL` please

libappstream-glib/as-store.c
```diff
@@ -3157,6 +3221,70 @@ as_store_load (AsStore *store,
 	return TRUE;
 }
 
+static void
+store_load_thread (GTask *task,
+		   gpointer source_object,
+		   gpointer task_data,
+		   GCancellable *cancellable)
+{
+	AsStore *store = source_object;
```
hughsie: Do we need a `AS_STORE()` here?

libappstream-glib/as-store.c
```diff
@@ -2879,6 +2879,70 @@ as_store_load_path (AsStore *store, const gchar *path,
...
+ * as_store_load_path_finish:
+ * @store: a #AsStore instance.
+ * @result: A #GAsyncResult
+ * @error: A #GError or %NULL.
+ *
+ * Retrieve the result of as_store_load_path_async().
+ *
+ * Returns: %TRUE for success
+ *
+ * Since: 0.7.2
```
hughsie: 0.7.7 now, sorry!

hughsie:
Sorry, I dropped the ball on this. A few tiny nits, and then LGTM.

libappstream-glib/as-store.c
fmuellner: Good question. (void*) pointers don't require a cast and this function is only called internally, so after adding g_return(_val)_if_fail to the public methods, we should be safe ... unless the checks are disabled via G_DISABLE_CHECKS.

Do you want me to add the cast for that case?

fmuellner: Sure, I added it now.

hughsie:
Fixed up the Since lines, and pushed, thanks.

PR #178, opened 2017-07-10
Support CompulsoryForDesktops in YAML

### LINKED ISSUE #175: Does not parse YAML CompulsoryForDesktops field
See https://appstream.debian.org/sid/main/metainfo/gnome-calculator.html for an example of this field in use and https://bugs.launchpad.net/ubuntu/+source/gnome-software/+bug/1698961 for the corresponding downstream bug report.

iainlane:
So that software centres don't present them as removable.

Closes #175


iainlane:
as requested on IRC!

Issue #181, opened 2017-07-24
Use Travis CI for build testing


williamcroberts:
One should add a build/test steps for the free travis CI: https://travis-ci.org/ and add the build badge to your README.

williamcroberts:
Are there tests for this?

ignatenkobrain:
so file pull request, no?

williamcroberts:
@ignatenkobrain You want me to create your Travis Account for you? This is something the project maintainer would need to do.

ignatenkobrain:
no,  want you to send .travis.yml ;)

ignatenkobrain:
hmm, looks like it is already there.

ignatenkobrain:
And actually it even works (https://github.com/hughsie/appstream-glib/pull/176). So closing.

williamcroberts:
@ignatenkobrain, you closed it without actually updating the README with a link to the travis ci. I have searched the public travis and see nothing. It's about providing testing and CI proof to consumers of the repo.

pohly:
@williamcroberts please see also https://github.com/hughsie/appstream-glib/pull/182#issuecomment-318037197. I think we can trust @hughsie that he is really doing the checks that he mentions there, so I am not sure whether we need further proofs (which might be hard to provide in public).

@hughsie if you do the checks in public (for example, on https://travis-ci.org/ or on the public Coverity scan service), then I suppose adding links would be useful.

williamcroberts:
If there is a test suite already there, then what can't be done in the public with respect to running it on the public travis? Coverity would definitely be nice. It's about test evidence, not about trust. I'm sure @hughsie is a great person, but proof is always nice.

hughsie:
> Coverity would definitely be nice

Well, this is the link, but I think you need a Coverity account to look at it, and you need permission to actually look at the scan results... https://scan.coverity.com/projects/hughsie-appstream-glib?tab=overview -- clang is done locally, I guess I could do this on travis too in the future.

> I'm sure @hughsie is a great person

Sure, I think so! :)

pohly:
On Wed, 2017-07-26 at 19:40 +0000, William Roberts wrote:
> Coverity would definitely be nice. It's about test evidence, not
> about trust.

Let's just say I know about a certain company which uses a certain
static code analysis tool and cannot talk about that in public.
Consumers of the open source projects published by that company have to
have trust in the company, because there are is no public test
evidence.
It's of course better if there is such evidence, but sometimes one has
[...]

PR #182, opened 2017-07-25
simplify build

pohly:
I'm working on packaging fwupd for OpenEmbedded, or more specifically,
https://github.com/intel/intel-iot-refkit.

OE currently has technical problems with gobject-introspection when
using meson. This is going to be addressed eventually, but some
intermediate solution is needed. As gir data is not needed for fwupd,
it would be great if it could be made optional. That might also be
useful in other builds that try to build only what's really needed.

I also found that the dependency checking was a bit too broad and
always pulled in gdk.


meson.build
```diff
@@ -78,6 +77,7 @@ if get_option('enable-builder')
     conf.set('HAVE_ALPM', 1)
   endif
   if get_option('enable-fonts')
+    gdk = dependency('gdk-3.0')
```
hughsie: Makes sense to me, thanks for catching this.

meson_options.txt
```diff
@@ -6,3 +6,4 @@ option('enable-fonts', type : 'boolean', value : true, description : 'enable fon
 option('enable-stemmer', type : 'boolean', value : true, description : 'enable stemmer support')
 option('enable-man', type : 'boolean', value : true, description : 'generate man pages')
 option('enable-gtk-doc', type : 'boolean', value : false, description : 'generate API reference')
+option('enable-gir', type : 'boolean', value : true, description : 'generate GObject Introspection data')
```
hughsie: I don't think enable-gir is right, it's disabling the .gir and the .typelib. I think enable-gobject-introspection would be much more accurate, although I concede it's a bit of a mouthful.

libappstream-glib/meson.build
```diff
@@ -212,19 +212,21 @@ introspection_sources = [
...
-  dependencies : deps,
-  includes : [
-    'GObject-2.0',
-    'Gio-2.0',
-    'GdkPixbuf-2.0'
-  ],
-  link_with : asglib,
-  install : true,
-)
+if get_option('enable-gir')
```
hughsie: If you're doing this here you also need to do it in libappstream-builder, even if you're not using it.

meson_options.txt
ignatenkobrain: I would call it, `enable-introspection`

pohly: enable-introspection is indeed better. It's also what the option is called in some autotools-based projects.

libappstream-glib/meson.build
pohly: Okay.

meson_options.txt
hughsie: Works for me, thanks!

hughsie: Maybe an obvious comment; but can they match this in fwupd too please.

pohly:
I've pushed revised patches. But before you merge, please let me also do some more testing locally. So far I've only made change without testing the resulting patches with OE.

hughsie:
Okay, both patches look fine to me, so yell when you've completed testing. I'd also be interested in any other "pain points" you experienced when trying to import fwupd.

pohly:
Patches are fine, also for fwupd.

Regarding pain points: the number of "new" libraries was a bit of work (libgusb, gcab). Besides providing the recipes, I also need to get the new components reviewed with regards to software quality, release processes, etc. You can help with that by clearly documenting whether you do automatic builds, static code analysis, etc. https://github.com/hughsie/appstream-glib/issues/181 is related to that.

I currently have no hardware to test it with - not much you can do about that, though ;-}

The dependency on polkit is problematic in an embedded device. It's overkill and drags in a JavaScript engine. The current solution is to use the polkit client library (so no changes needed in your code) with a pokit replacement service that implements the polkit D-Bus interface (see https://github.com/ostroproject/groupcheck).

Finally, some mechanism for feeding updates into fwupd without devices polling for updates would be useful. I'll file an enhancement request for that.

hughsie:
> Patches are fine, also for fwupd.

Can you push the branch for fwupd with the new parameter name please.

> You can help with that by clearly documenting whether you do automatic builds, static code analysis

Right, I'm open for ideas here. I'm slowly converting all my projects to using Travis-CI for each commit, but each tarball release is checked using clang, coverity and valgrind -- as all of these components are in RHEL. The review process for RHEL is arduous but does mean the code is presumably high quality.

I don't know a formal way of saying all that, so ideas welcome.

> I currently have no hardware to test it with - not much you can do about that, though ;-}

If you send me an email with your physical address I can send you a free PCB (it's really a ColorHug-v1 that failed QA and is unrepairable) that pretends it's a DFU device. I use one of those for the hardware.py checks that test downloading, deploying and downgrading a real hardware device.

> The dependency on polkit is problematic in an embedded device.

Any other ideas? Compiling out PolicyKit means that any user can do anything, which for an embedded device maybe isn't the end of the world.

> Finally, some mechanism for feeding updates into fwupd without devices polling for updates would be useful.

/me is suitably intrigued. I'll wait for the issue and discuss there.

pohly:
> Can you push the branch for fwupd with the new parameter name please.

Sorry, I had forgotten that. Done.

> If you send me an email with your physical address I can send you a
> free PCB (it's really a ColorHug-v1 that failed QA and is
> unrepairable) that pretends it's a DFU device. I use one of those for
> the hardware.py checks that test downloading, deploying and
> downgrading a real hardware device.

Thanks for the offer, but we'd also need other units for the automated
QA setup that we use for refkit. I need to think about this some more.

> Any other ideas? 
> Compiling out PolicyKit means that any user can do
> anything, which for an embedded device maybe isn't the end of the
> world.

No good idea either. I had considered making the polkit calls optional,
but even embedded devices may run less-trusted software which has
access to the system bus, so some kind of permission check is useful. I
think the current solution with groupcheck is good enough.

> /me is suitably intrigued. I'll wait for the issue and discuss there.

See https://github.com/hughsie/fwupd/issues/162

hughsie:
> but we'd also need other units for the automated QA setup that we use for refkit

There are currently 12 PCBs in the same drawer sitting there unloved, I don't know how many you need. The other way of testing this is just buying a few dozen Logitech unifying dongles -- which are also part of the pre-release tests I do. https://github.com/hughsie/fwupd/blob/master/data/installed-tests/hardware.py#L88

pohly:
On Wed, 2017-07-26 at 12:47 +0000, Richard Hughes wrote:
> The other way of testing this is just
> buying a few dozen Logitech unifying dongles -- which are also part
> of the pre-release tests I do.
> https://github.com/hughsie/fwupd/blob/master/data/installed-
> tests/hardware.py#L88

That looks like a usable approach. Just for my understanding, in that
test you are downgrading and then upgrading each time you run the test?

hughsie:
> you are downgrading and then upgrading each time you run the test

Yes, so it's only going to work a few (tens of?) thousand times before the dongle's EEPROM dies. Basically it's okay to do on each release, but probably not for each commit.

PR #191, opened 2017-09-18
Let appstream-builder emit new <launchables> 

mvollmer:
This pull request addresses the following issues:

- Appstream-builder throws away the type of launchables that it doesn't understand.  This PR makes it understand the new types from the 0.11 series of he AppStream spec. 

- Launchables don't appear in the appstream-builder output at all since it restricts itself to version 0.8 of the AppStream spec.  This PR makes it output according to version 0.11.4.

- The AppStream version is represented as a double, which has trouble working with anything that has a minor version above 0.9.  This PR switches to strings to represent versions.

### DIFF
--- a/libappstream-glib/as-launchable.c
+++ b/libappstream-glib/as-launchable.c
@@ -90,6 +90,10 @@ as_launchable_kind_from_string (const gchar *kind)
 {
 	if (g_strcmp0 (kind, "desktop-id") == 0)
 		return AS_LAUNCHABLE_KIND_DESKTOP_ID;
+	if (g_strcmp0 (kind, "service") == 0)
+		return AS_LAUNCHABLE_KIND_SERVICE;
+	if (g_strcmp0 (kind, "cockpit-manifest") == 0)
+		return AS_LAUNCHABLE_KIND_COCKPIT_MANIFEST;
 	return AS_LAUNCHABLE_KIND_UNKNOWN;
 }
 
@@ -108,6 +112,10 @@ as_launchable_kind_to_string (AsLaunchableKind kind)
 {
 	if (kind == AS_LAUNCHABLE_KIND_DESKTOP_ID)
 		return "desktop-id";
+	if (kind == AS_LAUNCHABLE_KIND_SERVICE)
+		return "service";
+	if (kind == AS_LAUNCHABLE_KIND_COCKPIT_MANIFEST)
+		return "cockpit-manifest";
 	return NULL;
 }
 
--- a/libappstream-glib/as-launchable.h
+++ b/libappstream-glib/as-launchable.h
@@ -57,6 +57,8 @@ struct _AsLaunchableClass
 typedef enum {
 	AS_LAUNCHABLE_KIND_UNKNOWN,
 	AS_LAUNCHABLE_KIND_DESKTOP_ID,		/* Since: 0.6.13 */
+	AS_LAUNCHABLE_KIND_SERVICE,		/* Since: 0.11.2 */
+	AS_LAUNCHABLE_KIND_COCKPIT_MANIFEST,    /* Since: 0.11.4 */
 	/*< private >*/
 	AS_LAUNCHABLE_KIND_LAST
 } AsLaunchableKind;


libappstream-glib/as-utils.c
```diff
@@ -2015,3 +2015,38 @@ as_utils_appstream_id_valid (const gchar *str)
...
+	minor += 1;
+
+	// A floating point number can only correctly represent a
+	// minor version that is a single digit.  Anything else is
+	// approximated by 0.9.  Anything after the minor version is
+	// ignored.
+	//
+	// For example, the string "0.7.5" is returned as 0.7, while
+	// "0.11" is returned as 0.9.
+
```
hughsie: Rather than doing this, can't we just mandate that appstream versions have three components, e.g. 0.6.0, 0.11.4 or 1.0.0?

mvollmer: Can you elaborate a bit?  What do refer to with "doing this"?  What does mandating three components change?  Does it simplify the code in `as_utils_version_to_double`?

libappstream-builder/asb-context.c
```diff
@@ -349,6 +371,23 @@ gdouble
...
+
+/**
+ * asb_context_get_api_version_string:
+ * @ctx: A #AsbContext
+ *
+ * Gets the target metadata API version.
+ *
+ * Returns: string
+ *
+ * Since: 0.1.0
```
hughsie: 0.7.3 surely?

libappstream-glib/as-icon.c
```diff
@@ -513,11 +513,12 @@ as_icon_node_insert_embedded (AsIcon *icon, GNode *parent, AsNodeContext *ctx)
 	n = as_node_insert (parent, "icon", NULL, 0,
 			    "type", as_icon_kind_to_string (priv->kind),
 			    NULL);
-	if (as_node_context_get_version (ctx) >= 0.8) {
+	if (as_node_context_version_at_least (ctx, "0.8")) {
 		as_node_add_attribute_as_uint (n, "width", priv->width);
 		as_node_add_attribute_as_uint (n, "height", priv->height);
 	}
-	if (priv->scale > 1 && as_node_context_get_version (ctx) > 0.9)
+	// XXX - figure out real minimum version for "scale"
```
hughsie: Maybe @ximion  knows?

libappstream-glib/as-utils.c
hughsie: Ahh, it's only used for the now-deprecated function. Gotcha.

libappstream-glib/as-utils.h
```diff
@@ -133,6 +133,7 @@ gchar		*as_utils_version_from_uint32	(guint32	 val,
 gchar		*as_utils_version_from_uint16	(guint16	 val,
 						 AsVersionParseFlag flags);
 gchar		*as_utils_version_parse		(const gchar	*version);
+gdouble          as_utils_version_to_double     (const gchar *version);
```
hughsie: `as-utils-private.h` surely?

libappstream-glib/as-icon.c
ximion: The minimal `scale` is 1, the maximum is undefined (but currently nothing gets bigger than 2x). I wanted to point you at the spec here, but it looks like this feature was implemented in AppStream without documenting it. Eww. (I'll fix that later)

libappstream-glib/as-utils.c
ximion: The libappstream library uses an enum for this, so I don't have to parse a string or do weird version comparisons. Maybe worth considering here as well.
See https://github.com/ximion/appstream/blob/master/src/as-metadata.h#L90

hughsie: An enum (that enum?) works for me.

libappstream-glib/as-icon.c
mvollmer: I am looking for the version of the AppStream spec that introduced the "scale" attribute for icons of type "embedded".  I couldn't quickly find that icon type in the spec, is it maybe not yet specified at all?

ximion: `embedded` is a proprietary extension that is AFAIK only used by OpenSUSE. I will never ever allow it in the spec.

libappstream-glib/as-utils.c
mvollmer: I have no opinion and I am happy to change the code to use an enum. @hughsie, can you make the call?

libappstream-builder/asb-context.c
mvollmer: Yes, fixed.

libappstream-glib/as-utils.h
mvollmer: Okay, fixed.

libappstream-glib/as-icon.c
mvollmer: Okay, how do we control whether appstream-builder outputs `embedded`?  Should the version enum have a AS_FORMAT_EXPERIMENTAL value?

mvollmer:
Okay, since a1481d201d089f9e70a0cf48a0424315e0f6cdaf we don't need to worry about versions anymore, so this gets a lot simpler.

PR #192, opened 2017-09-19
Allow icons in metainfo and appdata inputs

mvollmer:
And convert local icons to cached ones.


mvollmer:
See https://github.com/ximion/appstream/pull/138 for the corresponding spec addition.

mvollmer:
I noticed that `appstream-builder --enable-hidpi` produces XML like this:

```
<icon type="cached" height="64" width="64">64x64/cockpit.png</icon>
```

Note the unwanted "64x64" in the value of the icon tag.

This happens with `libappstream-glib-builder-0.7.2-1.fc26.x86_64` when processing the `cockpit-150-1.fc26.x86_64.rpm` package.

The actual collection data in `appstream-data-26-15.fc26.noarch` contains correct `<icon>` values for cockpit.destop.

Any idea what's going on?

libappstream-builder/plugins/asb-plugin-appdata.c
```diff
@@ -57,16 +60,102 @@ asb_plugin_check_filename (AsbPlugin *plugin, const gchar *filename)
...
+			  GError **error)
+{
+	AsIcon *icon;
+	GPtrArray *icons;
+	guint i;
+	guint min_icon_size;
+	g_autoptr(GdkPixbuf) pixbuf = NULL;
+	g_autofree gchar *name = NULL;
+
+	/* Convert local icons to cached.
```
hughsie: one line please

libappstream-builder/plugins/asb-plugin-appdata.c
```diff
@@ -57,16 +60,102 @@ asb_plugin_check_filename (AsbPlugin *plugin, const gchar *filename)
...
+	 */
+
+	icons = as_app_get_icons (AS_APP (app));
+	for (i = 0; i < icons->len; i++) {
```
hughsie: C99 please, i.e. `guint i = 0`

libappstream-builder/plugins/asb-plugin-appdata.c
```diff
@@ -57,16 +60,102 @@ asb_plugin_check_filename (AsbPlugin *plugin, const gchar *filename)
...
+		icon = g_ptr_array_index (icons, i);
+		if (as_icon_get_kind (icon) == AS_ICON_KIND_LOCAL) {
```
hughsie: What about the other icon kinds? e.g. `REMOTE` or `STOCK`? Shouldn't we add (without converting) those too?

hughsie:
> Any idea what's going on?

It's hugely and overly complicated. AppStream used to only support one icon size, and then we added HiDPI support which was optional. We actually changed the way we handled the HiDPI icons I think 3 times, and appstream-builder had to support them all for various versions of RHEL and Fedora. :( I think the builder we have configured for Fedora 27 produces:

    <icon type="cached" height="128" width="128">2dslamdemo.png</icon>
    <icon type="cached" height="64" width="64">2dslamdemo.png</icon>

mvollmer:
```
<icon type="cached" height="128" width="128">2dslamdemo.png</icon>
<icon type="cached" height="64" width="64">2dslamdemo.png</icon>
```

Isn't actually incorrect according to the spec?  Cached icons shouldn't tell their size, and the client is supposed to look through all cached locations to find the available sizes.  So, shouldn't the builder generate simply this:

```
<icon type="cached" >2dslamdemo.png</icon>
```

libappstream-builder/plugins/asb-plugin-appdata.c
mvollmer: The code modifies the existing icons, so any non-local ones are left in place.

We could filter out (or reject) cached icons, though.

ximion:
> Isn't that actually incorrect according to the spec? Cached icons shouldn't tell their size, and the client is supposed to look through all cached locations to find the available sizes. So, shouldn't the builder generate simply this:

Yes, this was incorrect - but @hughsie and some benchmarking convinced me that listing the sizes there is worth it due to the performance boost we get when knowing which sizes are available.
Since AppStream does not actually allow extending the amount of icons (unlike the themes approach the XDG icon theme standard establishes), encoding icon sizes does not cause any unwanted inflexibility as well.

mvollmer:
Now both local and stock icons are converted to remote.

mvollmer:
@hughsie, is there anything I can do to move this forward?  Or is the whole idea fundamentally wrong?

hughsie:
@mvollmer I think you've got the right idea, I'm just not sure if it's the right way to put it all in `asb-plugin-appdata.c` I think splitting this up so that we have something like `asb-plugin-icon.c` and using that to convert the local icon to the cached icon type would be much better. That way the `-desktop.c` plugin focuses on reading in the AsApp, and the `-appdata.c` plugin focuses on the XML. I can have a go at this, but do you have an .rpm that has an metainfo.xml file specifying an icon?

ximion:
FWIW, appstream-generator deals with stock/local icons exactly this way, creating cached icons from them.

hughsie:
@mvollmer does https://github.com/hughsie/appstream-glib/tree/wip/hughsie/icons-in-metainfo do what you expect?

mvollmer:
> does https://github.com/hughsie/appstream-glib/tree/wip/hughsie/icons-in-metainfo do what you expect?

Yes!

Sorry for missing the updates in this PR, and thanks a lot for picking this up.  I'll try out your branch and make sure the two existing Cockpit extensions work with it.

mvollmer:
> I'll try out your branch and make sure the two existing Cockpit extensions work with it.

Works!  

 - `g_error ("using stock")` of course crashes the program, but I guess you are not finished with that part, right?

 - Local icons should have a absolute filename as their value, right?  The `driver` test package doesn't have that.

mvollmer:
@hughsie, so I obviously want your branch to be merged. :-) Can I help?

hughsie:
@mvollmer I've just rebased my branch against master, and just waiting for CI to complete. If that works, I'll merge to master in a few minutes.

mvollmer:
> I merged my branch, thanks everybody! Yell if you find any problems.

Awesome, thanks!

PR #195, opened 2017-10-18
Add the WEB_APP URL kind

dchristidis:
The Telegram web application opens its [homepage](https://telegram.org/). Instead, it should open [this](https://web.telegram.org/) location. As it stands, a user has to perform unnecessary clicks each time the application is launched. Another example is Skype ([homepage](https://www.skype.com), [web app](https://web.skype.com)), which would be a wonderful addition once WebKitGTK+ can support it.

I suggest adding a new, optional type `AS_URL_KIND_WEB_APP`. If this is acceptable, I will submit a patch to GNOME Software so that it prefers it over `AS_URL_KIND_HOMEPAGE` during the web application installation.

### DIFF
--- a/libappstream-glib/as-enums.c
+++ b/libappstream-glib/as-enums.c
@@ -94,6 +94,8 @@ as_url_kind_to_string (AsUrlKind url_kind)
 		return "missing";
 	if (url_kind == AS_URL_KIND_TRANSLATE)
 		return "translate";
+	if (url_kind == AS_URL_KIND_WEB_APP)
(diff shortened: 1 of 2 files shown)


hughsie:
@ximion ?

dchristidis:
I was mistaken, I should have opened a PR on Appstream first. I will do so now.

dchristidis:
I’m closing this PR for now and I’ll open a new one once the next version of the AppStream specification is released.

PR #198, opened 2017-10-31
Add the 'hardware' requires type

hughsie:


superm1:
LGTM

ximion:
@hughsie This requires stuff isn't present in the spec - what is it used for any why does AppStream apparently have dependency logic now?

hughsie:
@ximion we're using it for firmware updating stuff. Maybe when it's working I'll propose it upstream, although I can't see anything other than fwupd using it.

ximion:
As long as you're not using those things in a custom tag, it's part of the spec though and should be in there.
What I am worried about (slightly) is that as soon as you propose it we'll change it to make it more generic or fit some broader scheme (for example, we use "modalias" in a `<provides/>`, tag, so a requires should probably use similar semantics?), and at that point we'll have yet another divergence.

hughsie:
Sure, I could see `modalias` being useful too, but that's a different thing to a CHID. I'm still in the "is this useful" stage before I propose it upstream, and when I do I'm happy to make changes and compromises like the other stuff I've done.

ximion:
> I'm still in the "is this useful" stage before I propose it upstream, and when I do I'm happy to make changes and compromises like the other stuff I've done.

That's perfectly fine then :-)
Until just now, I didn't know that something like a CHID even exists :P

superm1:
@ximion relevant details: https://docs.microsoft.com/en-us/windows-hardware/drivers/dashboard/using-chids and https://docs.microsoft.com/en-us/windows-hardware/drivers/dashboard/publish-a-driver-to-windows-update

Issue #201, opened 2017-11-09
Update list of depencies & build instructions


suve:
The list of depencies is outdated. On my system, I was missing the following:
```
gcab gperf json-glibc-devel libgcab-devel libstemmer-devel libuuid-devel meson
```
Out of these, `libgcab` being missing was not detected during `./configure`.

The build instructions are also outdated.

hughsie:
Could you do a PR for me against README.md please? I'd appreciate it very much.

PR #212, opened 2017-12-10
Update build instructions

suve:
This closes issue #201.

I've removed the `required: false` qualifier from `libgcab` because when I tried building `libappstream-glib` on my system, missing `libgcab` caused the build to fail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,11 +25,12 @@ To install the libappstream-glib library you either need to install the
 `libappstream-glib` package from your distributor, or you can build a local
 copy. To do the latter just do:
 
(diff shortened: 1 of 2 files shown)


hughsie:
Isn't libcab only required for the builder?

suve:
I checked again by removing `libgcab1-devel` from my system and running `./configure && make`.
```
../libappstream-glib/as-store-cab.c:24:10: fatal error: libgcab.h: No such file or directory
```

hughsie:
I think with the port to meson this PR is no longer valid. If it is, please re-open with a rebased patch. Thanks.

Issue #200, opened 2017-11-04
Error 'url-not-found' even though image exists


suve:
I'm currently packaging [cool-retro-term](https://github.com/Swordfish90/cool-retro-term) and I've encountered an issue with `appstream-util validate-relax`. I'm using the appdata file found in the repo, with minor changes (I removed the `width=""` and `height=""` from `<image>` since they do not match).

Anyways, the screenshots are defined as follows:
```
  <screenshots>
    <screenshot type="default">
      <caption>Default amber look</caption>
      <image>https://camo.githubusercontent.com/2443e662e95733ba6ae331f391f6ec036d1ee7fd/687474703a2f2f692e696d6775722e636f6d2f4e5566766e6c752e706e67</image>
    </screenshot>
    <screenshot>
      <caption>Apple II look</caption>
      <image>https://camo.githubusercontent.com/44a19842d532555c7b02bf6b4b4684add9edf18c/687474703a2f2f692e696d6775722e636f6d2f4d4d6d4d3648742e706e67</image>
    </screenshot>
  </screenshots>
```

Even though the images can be viewed in a browser, or fetched with `wget` or `curl` just fine, `appstream-util validate-relax` says:
```
FAILED:
? url-not-found         : <screenshot> url not found [https://camo.githubusercontent.com/44a19842d532555c7b02bf6b4b4684add9edf18c/687474703a2f2f692e696d6775722e636f6d2f4d4d6d4d3648742e706e67]
? url-not-found         : <screenshot> url not found [https://camo.githubusercontent.com/2443e662e95733ba6ae331f391f6ec036d1ee7fd/687474703a2f2f692e696d6775722e636f6d2f4e5566766e6c752e706e67]
```
One thing to note is that the file passed validation using version `0.6.13` of `libappstream-glib`, so the bug must have been introduced later.

suve:
It just occurred to me that the packaging guidelines in Fedora say to use the `--nonet` flag, so I guess it's actually a matter of the package builders being limited to their LAN (which makes sense, as a way to prevent downloading stuff during the package build). So this is most likely a false flag. Sorry for that.

Still, I wonder - maybe it'd make sense to change the error code to something like `url-not-reachable`? Personally, `url-not-found` reminds me of HTTP 404, which is an entirely different situation from "cannot reach server".

hughsie:
[...]

PR #214, opened 2017-12-17
Describe connection problems/HTTP errors when validating images

suve:
This changeset makes the image validation code display some descriptive error messages when downloading the image fails. Instead of a simple _"url not found"_ message, the user either gets a _"failed to connect message"_ or an HTTP error code. This should make it easier for users to differentiate between faulty networks, unresponsive or badly configured servers.

The `soup_status_get_phrase()` function is available since `libsoup-2.4`. ([link](https://git.gnome.org/browse/libsoup/tree/libsoup/soup-status.h#n96))

This fixes issue #200.

### DIFF
--- a/libappstream-glib/as-app-validate.c
+++ b/libappstream-glib/as-app-validate.c
@@ -24,6 +24,7 @@
 #include <fnmatch.h>
 #include <gdk-pixbuf/gdk-pixbuf.h>
 #include <libsoup/soup.h>
+#include <libsoup/soup-status.h>
 #include <string.h>
 
 #include "as-app-private.h"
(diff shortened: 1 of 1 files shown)


libappstream-glib/as-app-validate.c
```diff
@@ -471,10 +472,15 @@ ai_app_validate_image_check (AsImage *im, AsAppValidateHelper *helper)
...
+	if (SOUP_STATUS_IS_TRANSPORT_ERROR(status_code)) {
 		ai_app_validate_add (helper,
 				     AS_PROBLEM_KIND_URL_NOT_FOUND,
-				     "<screenshot> url not found [%s]", url);
+				     "<screenshot> failed to connect: %s [%s] ", soup_status_get_phrase(status_code), url);
+		return FALSE;
+	} else if (status_code != SOUP_STATUS_OK) {
+		ai_app_validate_add (helper,
+				     AS_PROBLEM_KIND_URL_NOT_FOUND,
+				     "<screenshot> HTTP %d: %s [%s]", status_code, soup_status_get_phrase(status_code), url);
```
hughsie: maybe something like `<screenshot> failed to download, HTTP error was %s [%d]`

Also, both added lines need wrapping to 80ish chars wide. Thanks.

suve:
To fit in the 80-char limit, I changed the indentation a bit. I wonder if the first argument to `ai_app_validate_add()` should't now go on a separate line, to be more consistent with the rest of the arguments.

hughsie:
If you merge that into one commit and force push to update the PR I think it's good to merge. Thanks.

suve:
So a squash, then? Sure.

suve:
Done.

PR #217, opened 2018-01-13
Fix build on musl systems.

maxice8:
Systems that use [musl](https://www.musl-libc.org/) as libc fail to compile with the following error

```
In file included from ../libappstream-glib/as-utils.c:37:0:
/usr/x86_64-linux-musl/usr/include/archive_entry.h:64:9: error: unknown type name ‘int64_t’
 typedef int64_t la_int64_t;
```

same happens with `client/as-utils.c`, including `stdint.h` before the `archive_entry.h` include fixes the build error.

Tested on Void Linux x86_64, i686, aarch64, armv7hf, x86_64-musl, armv6hf-musl, aarch64-musl. [Travis  Builds](https://travis-ci.org/maxice8/void-packages/builds/328361592)

### DIFF
--- a/client/as-util.c
+++ b/client/as-util.c
@@ -27,6 +27,7 @@
 #include <gio/gio.h>
 
 #include <appstream-glib.h>
+#include <stdint.h>
 #include <archive_entry.h>
 #include <archive.h>
 #include <libsoup/soup.h>
--- a/libappstream-glib/as-utils.c
+++ b/libappstream-glib/as-utils.c
@@ -34,6 +34,7 @@
 
 #include <fnmatch.h>
 #include <string.h>
+#include <stdint.h>
 #include <archive_entry.h>
 #include <archive.h>
 #include <libsoup/soup.h>


hughsie:
I don't think those patches are correct. I think the `stdint.h` include needs to go in the `archive_entry.h` header. appstream-glib doesn't use the `int64_t` type at all.

maxice8:
On Sat, Jan 13, 2018 at 12:07:42AM -0800, Richard Hughes wrote:
> I don't think those patches are correct. I think the `stdint.h` include needs to go in the `archive_entry.h` header. appstream-glib doesn't use the `int64_t` type at all.
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/hughsie/appstream-glib/pull/217#issuecomment-357417677

Yes i suspected that, archive_entry.h has nowhere where it gets stdint.h but apparently (my interpreation, take with a grain of salt, i'm not experienced) instead relies on archive.h being included first where it gets int64_t from a inttypes.h include https://github.com/libarchive/libarchive/blob/master/libarchive/archive.h#L56 , just flipping both the includes seems to fix for musl arches.

hughsie:
The include order shouldn't be important when using a library externally. I think this needs to be fixed in libarchive, sorry.

PR #220, opened 2018-01-18
Wip/hughsie/introspection fixes

hughsie:
Hey @kalev -- quick sanity check -- these changes don't break API/ABI, right?

### DIFF
--- a/libappstream-glib/as-app-validate.c
+++ b/libappstream-glib/as-app-validate.c
@@ -1135,7 +1135,7 @@ as_app_validate_check_id (AsAppValidateHelper *helper, const gchar *id)
  * Since: 0.1.4
  **/
 GPtrArray *
-as_app_validate (AsApp *app, AsAppValidateFlags flags, GError **error)
+as_app_validate (AsApp *app, guint32 flags, GError **error)
 {
 	AsAppProblems problems;
 	AsFormat *format;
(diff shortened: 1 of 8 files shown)


kalev:
I think the introspection fixes should be fine even if they end up changing the introspection API/ABI -- as far as I know, nothing in Fedora currently uses the appstream-glib introspection bindings.

As for the last patch that changes enums to guint16/guint32/guint64, my first guess would be that some of the changes there might change the C ABI, but not entirely sure. Could also be that they change the ABI on, let's say i686 and not on x86_64. Seems like a tricky thing to get right :)

kalev:
Also, if you are unsure about C ABI, I think it should be fine to bump soname. appstream-glib is basically gnome-software private library and I don't think anyone would yell if we bumped soname.

smcv:
You might be interested in https://bugzilla.gnome.org/show_bug.cgi?id=730932 which statically asserts some reasonable compiler behaviour, without which any vaguely recent GLib will fail to compile.

smcv:
It seems to be a pervasive assumption throughout the GLib/GNOME ecosystem that any reasonable compiler will ensure that all enums whose members are "fairly small" (fit in 31 bits) are the same size as an int. If a compiler didn't do this, strict ISO C would allow it, but most practical uses of GLib would either fail to work or break ABI regularly; hence the static assertion, to catch that early.

(In practice I don't think there's any reason anyone would implement their compiler or ABI that way - it would just break software for little or no benefit.)

hughsie:
> some of the changes there might change the C ABI

I was a little careful about this. I used a few tricks to force the enum to 16 bits (e.g. last entry of 0xffff) and also specifically typedef'ing the 64-bit one, so I think I'm fine on x64 at least.

hughsie:
@smcv much appreciated, thanks.

PR #229, opened 2018-03-20
for issues#228 update asb_package_guess_from_filename for inaccuracy rpm filename

CrazyCxl:
in appstream-glib/libappstream-builder/asb-package.c asb_package_guess_from_filename
---
if priv->filename likes `/xxx/xxx-xxx/xxx-xxx/free/x86_64/xxx.x86_64.rpm` , then priv will get
```
arch:x86_64
release:ccc/free/x86_64/test
version:bbb/ccc
name:bbb
```
Obviously, that's wrong.


hughsie:
I fixed this a slightly different way -- thanks for the PR tho.

PR #240, opened 2018-05-14
Do not parse firmware files

hughsie:
Firmware should now be distributed using the LVFS and consumed by fwupd.

This makes the library a lot simpler and drops a dep that most projects won't need.


hughsie:
@superm1 comments welcome

superm1:
Will this prevent being able to create a local appstream metadata to _use_ with fwupd via `appstream-util`?

hughsie:
Yes, although because we've been doing this on the LVFS it's now horribly out-of-specification. All the clever bits are in the LVFS, and I *think* it's okay to tell people to just spin up a private instance if they're not comfortable with the public one. Either way, it sucks having the code in two places, in different languages, handled in different ways.

superm1:
Well the main thing I'm thinking about is "local repository" support where an embedded distro may try to use a bunch of cabs that were locally built into the filesystem and generate the metadata from it.

I do get the problem with the same code in multiple languages but I think that there needs to be an easy way to generate that metadata without spinning up LVFS still. An easy to run python script is probably fine (maybe the LVFS code just repackaged into something easily executable).

PR #245, opened 2018-06-21
Add object type checks on the inputs of all functions

robert-ancell:


hughsie:
Do you really need the:

    AsAgreementSectionPrivate *priv;
    g_return_val_if_fail (AS_IS_AGREEMENT_SECTION (agreement_section), NULL);
    priv = GET_PRIVATE (agreement_section);

I thought GET_PRIVATE was just doing a pointer addition (i.e. not dereferencing anything) and was safe to do before the object check?

robert-ancell:
@hughsie I came to the same conclusion about 80% of the way through the patch, though I wasn't 100% sure and couldn't find any explicit documentation to assert that would be true. I'm happy to change them all to do it before the object check.

I thought there was a small risk of another variable definition using priv before the check, but this seems reasonable low (and should cease to exist when we finally start using C99 variable declarations at the point we need the variable...).

PR #247, opened 2018-06-26
Augment the max length of <li> tag to 300 in strict validation.

Jehan:
Our release `<description>` is not long, but clearly not one-worded. We still give a short description of the new features we list. Otherwise if we just drop titles, we may as well not write anything because that would be barely understandable.

Unfortunately it seems you limit list items to 100 characters. This is just too short IMO. It's like worse than a tweet or something! :P 

```
• style-invalid         : <li> is too long [Horizon Straightening: the Measurement tool is augmented with a button to rotate the active drawable with the measurement line used as horizon. It works on linked drawables as well.] maximum is 100 chars
```
Could we drop the limit or raise it to something reasonable (like at least 300 characters?).
Also nothing in the [spec](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-description) says anything either about size limitation.

### DIFF
--- a/libappstream-glib/as-app-validate.c
+++ b/libappstream-glib/as-app-validate.c
@@ -163,7 +163,7 @@ as_app_validate_description_li (const gchar *text, AsAppValidateHelper *helper)
 {
 	gboolean require_sentence_case = TRUE;
 	guint str_len;
-	guint length_li_max = 100;
+	guint length_li_max = 300;
 	guint length_li_min = 20;
 
 	/* relax the requirements a bit */


hughsie:
> Horizon Straightening: the Measurement tool is augmented with a button to rotate the active drawable with the measurement line used as horizon. It works on linked drawables as well.

This is too long. List elements are supposed to be one sentence long.

Jehan:
I think the problem is that we just have a very different definition of how should be the release notes. How would you reword so that the feature stays understandable within 100 chars?

For this example, I even chose an easy one. In the 2.10.4 list, I have longer items, for which a 100-char descriptions simple impossible to me. E.g.:

> PSD loader improvements: new file type "Photoshop image (merged)" to import a correctly rendered PSD file by loading pre-composited version of the image, even when some features are not supported by GIMP.

If we cannot write down understandable description of the release, I am just wondering what the feature is about (maybe we should stick to just give the release dates as most other projects are doing).

hughsie:
> PSD loader improvements: new file type "Photoshop image (merged)" to import a correctly rendered PSD file by loading pre-composited version of the image, even when some features are not supported by GIMP.

That's just too wordy. How about `Improve the PSD loader to import a pre-composited version of the image when not all features are supported`

> If we cannot write down understandable description of the release

The release notes are supposed to be just that, **notes** -- rather than a huge document used to promote the release. It's the same reason we don't allow images or screenshots in the release tags.

Jehan:
> Improve the PSD loader to import a pre-composited version of the image when not all features are supported

For the records, it's still too long. My editor counts 106 characters. So your version would also trigger an error with validate-strict. ;-)

> The release notes are supposed to be just that, notes -- rather than a huge document used to promote the release. It's the same reason we don't allow images or screenshots in the release tags.

Ok. Then maybe let's forget this request. What about discussing again (I think we discussed this as well, let's revive the discussion!) a release URL? For each of our releases, we have a corresponding blog post which does just that: longer descriptions, screenshots, etc. It's not necessarily that long nor exhaustive, but at least it's more than just a list of 100-chars lines.
But `validate-strict` also tells us:
```
<release> description should be prose and not contain hyperlinks
```

What about allowing hyperlinks, or simply allowing `<url/>` as possible child tag of `<release>` in appdata? Then people who want more info about a release can just click it.

Jehan:
Oh well I just found where we had this discussion. I completely forgot that I opened a pull request on the appstream repo long ago with a commit to amend the spec for exactly this:
https://github.com/ximion/appstream/pull/160

If we have this feature, I don't mind at all that descriptions could be just very small texts. Right now my problem is that we are stuck in this half-assed situation where appdata proposes us to write about our release while giving us no chance to give any details to people.

hughsie:
I think use `validate relax` and then I'll work on the appstream-glib and gnome-software functionality to display the URL in gnome-software.

Jehan:
Thanks! We are actually validating with relax currently, though I sometimes check what we would miss to be validated as "strict", hence the sudden flow of bug reports when I do such check. ;-)

Anyway looking forward to the possibility to add an URL to a release. :-)

hughsie:
> Anyway looking forward to the possibility to add an URL to a release.

We have this now: you can put a <url type="details"> into the release section.

PR #255, opened 2018-07-30
as-image: Handle images which are smaller than the destination

iainlane:
We currently don't consider this case at all, which results in negative
coordinates being passed to gdk_pixbuf_copy_area().

Let's instead look and see if the source image is smaller than the
destination. If it is, we just need to put it in the middle of our
desintation pixbuf.

This fixes a bug in gnome-software on Debian/Ubuntu where font previews are not shown, because they are smaller than the `1024x96` (?) that gnome-software is requesting and the code tries to upscale them. A later error makes this fail, but we should probably handle this by just never upscaling.


ximion:
+1 in general, but @hughsie has the last word.
One thing that leaves me puzzled here is that I experience this bug when running GNOME Software on GNOME Shell (no screenshots for fonts get shown), but when I run GS on KDE Plasma, the bug does *not* appear and GS works as expected.
So, for some reason GNOME Software works better on Plasma than on GNOME.
I also have no idea why the DE would have any impact on the behavior of a image scaling/padding function, so at the moment I just admire this bug for its weirdness.

Regardless, Iains patch looks good to me and makes sense.

robert-ancell:
This change seems overly complex - I think the solution is just to fix the scaling code to handle all requested sizes - see https://github.com/hughsie/appstream-glib/pull/256

iainlane:
Thanks Robert. The outcomes are different between the two cases though, I think. In this one we never upscale, but just position the source image in the middle, and in the other one (AFAICS) you upscale the required dimension so it fits.

This PR:
screenshot of this PR

#256:
screenshot of 256

robert-ancell:
@iainlane yeah, correct the other MR scales in all cases. Your method looks a lot sharper which seems the right solution here. I'm not sure what the intention of `as_image_save_pixbuf` is supposed to be - I guess we'll have to see what @hughsie says.

libappstream-glib/as-image.c
```diff
@@ -830,8 +839,8 @@ as_image_save_pixbuf (AsImage *image,
 			      (gint) tmp_width,
 			      (gint) tmp_height,
 			      pixbuf,
-			      (gint) (width - tmp_width) / 2,
-			      (gint) (height - tmp_height) / 2);
+			      ABS ((gint) (width - tmp_width)) / 2,
+			      ABS ((gint) (height - tmp_height)) / 2);
```
robert-ancell: This ABS shouldn't be required, right? Since the above code should always ensure tmp_width <= width and tmp_height <= height. If you added this for safety you could put some assertions before this.

hughsie: @iainlane?

iainlane: sorry @hughsie @robert-ancell I forgot to reply. That also means I now forgot what the code does... let's see...

OK, here we go I think. I've added `if (pixbuf_width > width || pixbuf_height > height)`, which wasn't there before. In the `else` part of that `if` - the case where padding is required - I set `tmp_width` and `tmp_height` to the image's width and height.

That means that `width ≥ tmp_width` and `height ≥ tmp_height`, so `width - tmp_width` is always positive.

So in summary, I think the `ABS` is not needed. I'll rebase & remove it. :+1:

iainlane:
Isn't the value that GNOME software uses essentially arbitrary? Not sure it would be right for the appstream generator to have knowledge of the value `1024×96` to be honest.

I think icons use `AsIcon` and this is essentially only for screenshots, or is that wrong?

ximion:
@kalev The AppStream spec is very explicit about icon sizes (there are rules that must be followed), but for screenshots there are only recommendations for apps and the software center has to deal with whatever it gets.
That also means though that there would be nothing wrong with having appstream-generator produce arbitrary screenshot sizes for fonts, including 1024x96px (if that doesn't mess up other software centers that started to rely on other sizes).

PR #256, opened 2018-08-02
as-image: Correctly scale and position images into requested size

robert-ancell:
The previous code didn't handle output dimensions that were needed letterboxing.


robert-ancell:
This is an alternative solution to https://github.com/hughsie/appstream-glib/pull/255

robert-ancell:
I'm pretty sure that #255 is more correct, so closing this.

Issue #253, opened 2018-07-27
appstream-util: inconsistent argument order between HTML functions


suve:
- `status-csv` and `status-html` expect the following arguments: `input.tar.gz output.{csv,html}`

- `matrix-html` expects the following arguments: `output.html input.tar.gz`

I think it'd be good to maintain some consistency regarding the ordering of arguments. 

Also, `matrix-html` can take multiple `input.tar.gz` arguments, while the `status-*` functions can only take one. Is there any particular reason behind this?

hughsie:
I guess `output` then `input1` [`inputn`] globally makes sense. Would you be able to submit a PR for this fix please?

suve:
Personally I was thinking about `input1 [input2...] output`, since that's the order `cp` and other coreutils use, but that's mostly a matter of preference.

I will submit a pull request in the following days.

hughsie:
Yup, `input1 [input2...] output` would be fine too for me.

PR #259, opened 2018-08-22
Change order of arguments in matrix-html to match status-html

suve:
This changeset fixes issue #253.

There's one thing I'm wondering about, though - `matrix-html` can take multiple input.tar.gz arguments, while the `status-*` functions can only take one. Is there any particular reason behind this? Because if not, then I think that also should be changed to make the API more consistent.

### DIFF
--- a/client/as-util.c
+++ b/client/as-util.c
@@ -2495,18 +2495,19 @@ as_util_matrix_html (AsUtilPrivate *priv, gchar **values, GError **error)
 	g_autoptr(GString) html = NULL;
 
 	/* check args */
-	if (g_strv_length (values) < 2) {
+	guint value_count = g_strv_length (values);
+	if (value_count < 2) {
 		g_set_error_literal (error,
 				     AS_ERROR,
 				     AS_ERROR_INVALID_ARGUMENTS,
 				     "Not enough arguments, "
-				     "expected matrix.html filename.xml.gz");
+				     "expected filename.xml.gz matrix.html");
 		return FALSE;
 	}
 
 	/* load file */
 	store = as_store_new ();
-	for (i = 1; values[i] != NULL; i++) {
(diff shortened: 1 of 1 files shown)


client/as-util.c
```diff
@@ -2495,18 +2495,19 @@ as_util_matrix_html (AsUtilPrivate *priv, gchar **values, GError **error)
 	g_autoptr(GString) html = NULL;
 
 	/* check args */
-	if (g_strv_length (values) < 2) {
+	guint value_count = g_strv_length(values);
```
hughsie: space before the bracket pls

client/as-util.c
```diff
@@ -2495,18 +2495,19 @@ as_util_matrix_html (AsUtilPrivate *priv, gchar **values, GError **error)
...
 				     "Not enough arguments, "
-				     "expected matrix.html filename.xml.gz");
+				     "expected filename.xml.gz matrix.html");
 		return FALSE;
 	}
 
 	/* load file */
 	store = as_store_new ();
-	for (i = 1; values[i] != NULL; i++) {
+	for (i = 0; i < value_count-1; i++) {
```
hughsie: space either side of the `-` pls.

client/as-util.c
```diff
@@ -2567,7 +2568,7 @@ as_util_matrix_html (AsUtilPrivate *priv, gchar **values, GError **error)
 	g_string_append (html, "</html>\n");
 
 	/* save file */
-	return g_file_set_contents (values[0], html->str, -1, error);
+	return g_file_set_contents (values[value_count-1], html->str, -1, error);
```
hughsie: same.

hughsie:
Mostly looks good to me, a few tiny nits. Thanks!

suve:
Done via amending the commit.

PR #261, opened 2018-08-29
Fix potential buffer underflow and inefficient copy using fnmatch.

robert-ancell:
If length was zero we could check the -1 index.
A nul was always added because we only used the length of the string, not the
buffer (i.e. off by one).
Also remove a check for a negative number from an unsigned number.


libappstream-glib/as-app.c
```diff
@@ -6323,7 +6325,7 @@ as_app_parse_appdata_file (AsApp *app,
 			     filename, error_local->message);
 		return FALSE;
 	}
-	data = g_bytes_new_take (g_steal_pointer (&data_raw), len);
+	data = g_bytes_new_take (g_steal_pointer (&data_raw), len + 1);
```
hughsie: This is wrong, this would read beyond the allocation.

libappstream-glib/as-app.c
```diff
@@ -6208,7 +6208,9 @@ as_app_parse_appdata_guess_project_group (AsApp *app)
 static int
 as_utils_fnmatch (const gchar *pattern, const gchar *text, gsize text_sz, gint flags)
 {
-	if (text_sz != -1 && text[text_sz-1] != '\0') {
+	if (text_sz == 0)
+		return FNM_NOMATCH;
+	if (text[text_sz-1] != '\0') {
```
hughsie: Why not just?

    	if (text_sz != 0 && text[text_sz-1] != '\0') {

libappstream-glib/as-app.c
hughsie: > and a nul is added to the end

I've learned something today... okay, so that hunk is good, but could you add a comment to help 5-year-old-me for the future?

hughsie:
CI seems to be failing -- can you check than out pls?

hughsie:
@robert-ancell ?

PR #267, opened 2018-10-25
as-app: Disable subsuming of releases with duplicate versions

superm1:
In the case of composite firmware packages it's possible that the same
firmware binary will be in two different stores but the newer store
is preferred since other binaries in the store are newer.

Fixes: fwupd #818


superm1:
Closing due to preference of a different approach to fix.

PR #268, opened 2018-10-30
as-store: Guarantee to return G_IO_ERROR_NOT_FOUND when loading fails

pwithnall:
Previously, as_store_from_file() would either return
G_IO_ERROR_NOT_FOUND or AS_STORE_ERROR_FAILED if the given file did not
exist, depending on its file extension. That is not particularly helpful
for callers.

Instead, guarantee to always return G_IO_ERROR_NOT_FOUND if the file was
not found.

### DIFF
--- a/libappstream-glib/as-store.c
+++ b/libappstream-glib/as-store.c
@@ -2017,7 +2017,10 @@ as_store_from_file_internal (AsStore *store,
 	if (priv->add_flags & AS_STORE_ADD_FLAG_ONLY_NATIVE_LANGS)
 		flags |= AS_NODE_FROM_XML_FLAG_ONLY_NATIVE_LANGS;
 	root = as_node_from_file (file, flags, cancellable, &error_local);
-	if (root == NULL) {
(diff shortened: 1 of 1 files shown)


hughsie:
I think the general idea was that we'd use AsStoreError domains where-ever possible. Shouldn't the `G_IO_ERROR_NOT_FOUND` be converted to be `AS_STORE_ERROR_FAILED` instead? I concede it's not hugely helpful.

pwithnall:
> I think the general idea was that we'd use AsStoreError domains where-ever possible.

Why? That either means you reduce the specificity of error reporting (by coalescing multiple error codes from other domains into a single `AS_STORE_ERROR_FAILED` code), or you duplicate loads of error codes from other domains.

> Shouldn't the `G_IO_ERROR_NOT_FOUND` be converted to be `AS_STORE_ERROR_FAILED` instead? I concede it's not hugely helpful.

The motivation behind me writing this patch was so that I could distinguish “file not found” errors from everything else when loading an AppStream file.

PR #269, opened 2018-11-01
as-content-rating: Add as_content_rating_get_rating_ids()

pwithnall:
This allows the set of rating IDs in the AsContentRating to be returned,
allowing the caller to iterate over them to check their values, rather
than having to hard-code their own list of OARS rating IDs.

Includes a basic test.

### DIFF
--- a/libappstream-glib/as-self-test.c
+++ b/libappstream-glib/as-self-test.c
@@ -1700,6 +1700,9 @@ as_test_content_rating_func (void)
 	gboolean ret;
 	g_autoptr(AsNodeContext) ctx = NULL;
 	g_autoptr(AsContentRating) content_rating = NULL;
+	g_autofree const gchar **rating_ids = NULL;
+	const gchar *expected_rating_ids[] = { "drugs-alcohol", "violence-cartoon", NULL };
+	gsize i;
 
 	content_rating = as_content_rating_new ();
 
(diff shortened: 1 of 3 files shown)


libappstream-glib/as-self-test.c
```diff
@@ -1722,6 +1725,15 @@ as_test_content_rating_func (void)
 			 AS_CONTENT_RATING_VALUE_MILD);
 	g_assert_cmpint (as_content_rating_get_value (content_rating, "violence-bloodshed"), ==,
 			 AS_CONTENT_RATING_VALUE_UNKNOWN);
+
+	rating_ids = as_content_rating_get_rating_ids (content_rating);
+	g_assert_nonnull (rating_ids);
+
+	for (i = 0; rating_ids[i] != NULL && expected_rating_ids[i] != NULL; i++)
```
hughsie: `for (gsize i = 0; ` pls

hughsie:
One little nit, and then LGTM, thanks. I've given you push access to master, so please just merge when you've fixed it up locally.

libappstream-glib/as-self-test.c
pwithnall: Not possible, since `i` is needed outside the loop to check that both lists have terminated (with `NULL`) at the same time.

pwithnall:
Thanks for the fast review!

libappstream-glib/as-self-test.c
hughsie: /me gets more coffee -- thanks.

PR #273, opened 2018-11-16
Search icons in $prefix/share even when $prefix/usr exists.

### LINKED ISSUE #271: appstream-compose fails to find icon when $prefix/usr exists
I have an appstream file, a desktop file and an icon used as exports in a flapkak application. The appstream file references the desktop file. The desktop file uses the icon.

When `$prefix/usr` exists, `appstream-compose` requires the icon files to be in `$prefix/usr/share/icons` instead of `$prefix/share/icons`.

Flatpak is the use case I've encountered, but it's reproducible without it.

----

```sh
~/appstream-bug$ tree
.
├── share
│   ├── applications
│   │   └── org.example.TestIcon.desktop
│   ├── icons
│   │   └── hicolor
│   │       └── 256x256
│   │           └── apps
│   │               └── org.example.TestIcon.png
│   └── metainfo
│       └── org.example.TestIcon.appdata.xml
└── usr

~/appstream-bug$ appstream-compose --verbose --prefix=$(pwd) org.example.TestIcon
WARNING: Metadata origin not set, using 'example'
Processing application org.example.TestIcon
(appstream-compose:29789): As-DEBUG: Looking for /home/bak/appstream-bug/share/metainfo/org.example.TestIcon.appdata.xml
Error loading desktop file: Failed to find icon: Failed to find icon org.example.TestIcon
```

  

*org.example.TestIcon.desktop*:
```ini
[Desktop Entry]
Name=Test icon bug
Exec=/bin/sh
Icon=org.example.TestIcon
Type=Application
```

*org.example.TestIcon.appdata.xml*:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<component type="desktop-application">
  <id>org.example.TestIcon</id>
  <metadata_license>CC-BY-SA-3.0</metadata_license>
  <description>
  </description>
  <launchable type="desktop-id">org.example.TestIcon.desktop</launchable>
</component>
```

BAKFR:
Fixes #271 

With this commit, `as_utils_find_icon_filename_full()` loop over both `$prefix/usr/share` and `$prefix/share`. If it doesn't find the icon in the first folder, it goes into the second instead of aborting the search. The case `prefix == NULL` still works as before.

Feel free to ask for clarification or for code changes.

### DIFF
--- a/libappstream-glib/as-utils.c
+++ b/libappstream-glib/as-utils.c
@@ -884,7 +884,9 @@ as_utils_find_icon_filename_full (const gchar *destdir,
 	guint j;
 	guint k;
 	guint m;
+	guint n;
 	const gchar **sizes;
+	const gchar *share_dirs[] = { "usr/share", "share", NULL };
 	const gchar *pixmap_dirs[] = { "pixmaps", "icons", NULL };
 	const gchar *theme_dirs[] = { "hicolor", "oxygen", NULL };
 	const gchar *supported_ext[] = { ".png",
@@ -923,7 +925,6 @@ as_utils_find_icon_filename_full (const gchar *destdir,
 				 "status",
 				 "stock",
 				 NULL };
-	g_autofree gchar *prefix = NULL;
 
 	g_return_val_if_fail (search != NULL, NULL);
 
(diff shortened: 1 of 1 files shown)


hughsie:
I'm a bit confused. Isn't the prefix supposed to contain the `usr` section too? The prefix is certainly a different thing to the destdir... https://www.gnu.org/prep/standards/html_node/Directory-Variables.html

If I understand correctly your code is changing the icon search to look in `prefix/share` and also `destdir/share`, rather than the old code that looked to see if `destdir/usr` existed, and if not used `destdir` as the prefix.

If that's right, I'm not sure we want to support both -- in fact, maybe we move the fallback out of `as_utils_find_icon_filename_full` and just call it twice with different destdirs.

BAKFR:
I'm not familiar with the GNU terminology. I've read your link, and indeed, what I've called $prefix in my commit message is the destdir.

From what I understand, the prefix would be an "usr" folder used to store things, usually `/usr` or `/usr/local`. The destdir would correspond to the root folder. In flatpak context, `/app` is the prefix used by applications, right ?

> If I understand correctly your code is changing the icon search to look in prefix/share and also destdir/share, rather than the old code that looked to see if destdir/usr existed, and if not used destdir as the prefix.

Basically, yes.
The behavior between both code is almost identical. The only change is when `destdir/usr` exists and the icon is not found: the old code will returns an error, while the new code will try to use destdir as the prefix anyway.
When `destdir/usr` doesn't exist, both code will use destdir as the prefix.

> If that's right, I'm not sure we want to support both -- in fact, maybe we move the fallback out of as_utils_find_icon_filename_full and just call it twice with different destdirs.

Currently `as_utils_find_icon_filename_full()` is sometimes called with a prefix (in `client/as-compose.c`) and sometimes with a destdir (in `libappstream-builder/plugins/asb-plugin-icon.c`). Hence the need for this fallback.

Do we need this fallback ? Can we not create two distinct functions ?

hughsie:
I don't think $prefix/usr makes sense, sorry.

PR #276, opened 2018-12-14
Various locking and thread safety fixes for AsStore

kalev:
This should fix a number of gnome-software crashes in F29 that usually involve one thread modifying AsStore, and the other one reading back from it.

It's not always that gnome-software spawns two worker threads (this we could fix by adding locking on gnome-software side); sometimes it's the AsStore internal file monitor watch callback firing in the main loop and then changing AsStore from the callback (which runs in a different thread from the plugin loader's worker threads).

### DIFF
--- a/libappstream-glib/as-store.c
+++ b/libappstream-glib/as-store.c
@@ -1,6 +1,7 @@
 /* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*-
  *
  * Copyright (C) 2013-2016 Richard Hughes <richard@hughsie.com>
+ * Copyright (C) 2015-2018 Kalev Lember <klember@redhat.com>
  *
  * Licensed under the GNU Lesser General Public License Version 2.1
  *
@@ -264,6 +265,22 @@ as_store_changed_uninhibit_cb (void *v)
 
 #define _cleanup_uninhibit_ __attribute__ ((cleanup(as_store_changed_uninhibit_cb)))
 
+static GPtrArray *
+_dup_app_array (GPtrArray *array)
+{
+	GPtrArray *array_dup;
+
+	g_return_val_if_fail (array != NULL, NULL);
+
+	array_dup = g_ptr_array_new_full (array->len, (GDestroyNotify) g_object_unref);
+	for (guint i = 0; i < array->len; i++) {
+		AsApp *app = g_ptr_array_index (array, i);
+		g_ptr_array_add (array_dup, g_object_ref (app));
+	}
+
+	return array_dup;
+}
+
 /**
  * as_store_add_filter:
  * @store: a #AsStore instance.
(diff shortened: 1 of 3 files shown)


libappstream-glib/as-store.c
```diff
@@ -264,6 +265,23 @@ as_store_changed_uninhibit_cb (void *v)
 
 #define _cleanup_uninhibit_ __attribute__ ((cleanup(as_store_changed_uninhibit_cb)))
 
+static GPtrArray *
+_dup_app_array (GPtrArray *array)
+{
+	GPtrArray *array_dup;
+
+	if (array == NULL)
+		return g_ptr_array_new_with_free_func ((GDestroyNotify) g_object_unref);
```
hughsie: Do we need these two lines? Does g_ptr_array_new_full explode in a ball of fire if you try to allocate it a size 0?

kalev: Setting it to 0 should be fine. We could do something like:

```
GPtrArray *array_dup = g_ptr_array_new_full (array != NULL ? array->len : 0, (GDestroyNotify) g_object_unref);`

if (array == NULL)
        return g_steal_pointer (&array_dup);

...
```

Would that be cleaner?

hughsie: Ohh! `array == NULL`, i read that as as `array->len == 0` for some reason. What you had was cleaner, ignore me. I do wonder if callers to this function should check for NULL tho....

kalev: I think it's easier to just return an empty container here because in all the public API that did "return g_object_ref (priv->array)" so far (and that I've changed to return a deep copy now) it was always returning an empty array and not NULL and clients probably rely on that now.

hughsie: Right, but wouldn't a `g_return_val_if_fail(array !=NULL)` be a sensible precaution instead?

libappstream-glib/as-store.c
```diff
@@ -272,8 +272,7 @@ _dup_app_array (GPtrArray *array)
 {
 	GPtrArray *array_dup;
 
-	if (array == NULL)
-		return g_ptr_array_new_with_free_func ((GDestroyNotify) g_object_unref);
+	g_assert (array != NULL);
```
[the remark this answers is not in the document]
kalev: Isn't the convention to use assert for internal functions where we control all the inputs (and check the return values in callers like I've changed it to do here), and g_return_val_if_fail in public API where a client might call stuff wrongly?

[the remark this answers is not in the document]
hughsie: I don't really use assert, as usually the public stuff ends up calling the "internal" stuff sooner or later.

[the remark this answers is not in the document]
kalev: Sure, let's do g_return_val_if_fail then :) Rebased and fixed up.

Issue #270, opened 2018-11-07
appstream-util doesn't compare versions correctly


scx:
It seems that appstream-util doesn't compare versions correctly, which leads to incorrect validation of AppData files.

```
$ appstream-util --version
Version:	0.7.13
```

```
$ appstream-util vercmp "Build 9half" "Build 10"
Build 9half > Build 10
$ appstream-util vercmp "Build 9.5" "Build 10"
Build 9.5 > Build 10
$ appstream-util vercmp "Build 9+" "Build 10"
Build 9+ > Build 10
$ appstream-util vercmp "Build 9a" "Build 10"
Build 9a > Build 10
$ appstream-util vercmp "Build 9" "Build 10"
Build 9 > Build 10
```

```
$ appstream-util vercmp "9half" "10"
9half > 10
$ appstream-util vercmp "9.5" "10"
9.5 > 10
$ appstream-util vercmp "9+" "10"
9+ > 10
$ appstream-util vercmp "9a" "10"
9a > 10
$ appstream-util vercmp "9" "10"
9 < 10
```

As you can see, only the last result is correct.
The problem occurs in both Freedesktop 1.6 and 18.08.

Please compare this with the rpmdev-vercmp on RHEL 7:

```
$ rpm -q --qf "%{VERSION}\n" rpmdevtools
8.3
```

```
$ rpmdev-vercmp "Build 9half" "Build 10"
Build 9half < Build 10
$ rpmdev-vercmp "Build 9.5" "Build 10"
Build 9.5 < Build 10
$ rpmdev-vercmp "Build 9+" "Build 10"
Build 9+ < Build 10
$ rpmdev-vercmp "Build 9a" "Build 10"
Build 9a < Build 10
$ rpmdev-vercmp "Build 9" "Build 10"
Build 9 < Build 10
```

```
$ rpmdev-vercmp "9half" "10"
9half < 10
$ rpmdev-vercmp "9.5" "10"
9.5 < 10
$ rpmdev-vercmp "9+" "10"
9+ < 10
$ rpmdev-vercmp "9a" "10"
9a < 10
$ rpmdev-vercmp "9" "10"
9 < 10
```

See also: https://github.com/widelands/widelands/blob/600993f5051752cf7af2a9d572da0ffbb01d8089/debian/widelands.appdata.xml#L265-L284

hughsie:
[...]

PR #279, opened 2019-01-05
Use rpmvercmp instead of internal vercmp when built with rpm support

kalev:
This should help make sure that the version comparison algorithm matches exactly with Fedora's when built on Fedora and supports caret, tilde, and plus characters that all have special semantics in Fedora packages.

https://github.com/hughsie/appstream-glib/issues/270

### DIFF
--- a/libappstream-glib/as-utils.c
+++ b/libappstream-glib/as-utils.c
@@ -40,6 +40,10 @@
 #include <stdlib.h>
 #include <uuid.h>
 
+#ifdef HAVE_RPM
+#include <rpm/rpmlib.h>
+#endif
+
 #include "as-app-private.h"
 #include "as-enums.h"
 #include "as-node.h"
@@ -1465,7 +1469,11 @@ as_utils_vercmp_full (const gchar *version_a,
 		g_autofree gchar *str_b = as_utils_version_parse (version_b);
 		return as_utils_vercmp_internal (str_a, str_b);
 	} else {
+#ifdef HAVE_RPM
+		return rpmvercmp (version_a, version_b);
+#else
 		return as_utils_vercmp_internal (version_a, version_b);
+#endif
 	}
 }
 
--- a/libappstream-glib/as-utils.c
+++ b/libappstream-glib/as-utils.c
@@ -1387,45 +1387,17 @@ as_utils_vercmp_chunk (const gchar *str1, const gchar *str2)
 	return as_utils_vercmp_char (str1[i], str2[i]);
 }
 
-/**
- * as_utils_vercmp_full:
- * @version_a: the release version, e.g. 1.2.3
- * @version_b: the release version, e.g. 1.2.3.1
- * @flags: some #AsVersionCompareFlag
- *
- * Compares version numbers for sorting.
- *
- * Returns: -1 if a < b, +1 if a > b, 0 if they are equal, and %G_MAXINT on error
- *
- * Since: 0.7.15
- */
-gint
-as_utils_vercmp_full (const gchar *version_a,
-		      const gchar *version_b,
-		      AsVersionCompareFlag flags)
+static gint
+as_utils_vercmp_internal (const gchar *version_a,
+                          const gchar *version_b)
 {
 	guint longest_split;
 	g_auto(GStrv) split_a = NULL;
 	g_auto(GStrv) split_b = NULL;
 
-	/* sanity check */
-	if (version_a == NULL || version_b == NULL)
-		return G_MAXINT;
+	split_a = g_strsplit (version_a, ".", -1);
+	split_b = g_strsplit (version_b, ".", -1);
 
-	/* optimisation */
-	if (g_strcmp0 (version_a, version_b) == 0)
-		return 0;
-
-	/* split into sections, and try to parse */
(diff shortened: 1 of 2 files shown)


ignatenkobrain:
Actually + doesn't have any special meaning, it's same as .

ximion:
Ultimately Richard's choice, but I would be concerned by a compile-flag changing the behavior of version comparison on some distributions, but not others in case the algorithm is ever adjusted in either RPM or appstream-glib.
By just always using the internal algorithm instead of the RPM one, at least the behavior would be consistent everywhere. AppStream is designed to be cross-distro, so (potentially) having a version comparison throw out a different result depending on where appstream-glib was built feels wrong.
(It would be neat to have the one version comparison that rules them all, but writing it is very hard (there are so many corner cases to take care of...))

hughsie:
I'm a little hesitant to depend on rpm, but I'm more concerned about rpmvercmp() and the internal vercmp() differing in results. Would we accept a patch to use the dpkg vercmp routine instead? Maybe it would be best to use rpmvercmp, and then fall back to an internal version of the rpmvercmp code.

kalev:
Just to make it clear, I'm not pushing much for this solution, just want to get something that works for Fedora packages. Copying the librpm implementation would be fine with me as well. When we chatted about this last year, @hughsie thought that maybe it would make sense to link to librpm instead so that's why the patch here.

So the constraints we have. On one hand, on the distro side we have:

 1) appstream data composed from packages, so we need to match the version comparison algorithm from packages;

and on the upstream side:

 2) appdata files written by upstream and we need to make sure the versions in <release> tags compare the same way in all distros.

These are somewhat conflicting goals, sadly. I don't have any good ideas how to solve both.

I think in practice the rpm and dpkg algorithms are fairly similar and if any upstream goes outside of the subset that's supported by both they are slightly crazy :)

Maybe the best way to solve this situation here is to make sure we have good test coverage and keep the appstream-glib and appstream tests in sync, so that the algorithms are known to be compatible? And make sure that the same tests pass with both rpmvercmp and with the internal vercmp implementation.

Also, I think if we go with the patch here, we are still going to have to improve the internal vercmp implementation along the lines suggested in https://github.com/hughsie/appstream-glib/issues/270 -- we can't leave the appstream-builder that runs in flatpak (and doesn't link with librpm) out in the cold.

ignatenkobrain:
Why can't we use semver,?

ximion:
@smcv 
> This is always a good idea, IMO. If there are indeed corner cases where RPM and dpkg differ, comments in the test data would be a great place to highlight these.

Jup, having comprehensive tests would be awesome. The AppStream testsuite has a few, but a lot more would be needed. I expect the differences to only affect few corner cases though.
The only reason why I used the RPM vercmp algorithm in AppStream (libappstream) to begin with was to have some generic way to compare versions. I originally wanted to use the dpkg algorithm, but the RPM one was more readable and compact. The only reason why I documented this was so people know about it if they wonder how versions are compared in the implementation, and *not* as a spec recommendation.

@ignatenkobrain Strict semver just doesn't work in the real world. I'd bet that only about 50% of version numbers in the Debian archive would fit that scheme, maybe even less.

@kalev My original intent was to not actually have AppStream compare version numbers provided by the package manager. For that purpose, a software management frontend should just call into PackageKit, which *for sure* will do the exact version comparison the package manager uses.
What the version comparison was intended for is things like Flatpaks and firmware.
I see a need to compare those version numbers to package version numbers, but for that I am actually planning a dedicated comparison function which handles epochs correctly depending on the source of a component (e.g. when comparing a Flatpak component to a package component, the epoch of the version number from the latter would be dropped. Otherwise, the package would always win). I didn't get around to implementing that yet though (features like this, I learned, need more thought before just throwing them in).

smcv:
> Why can't we use semver,?

semver is a very simplified versioning scheme. If upstreams want to use it then they're welcome to (any reasonable version comparison scheme should sort semver versions correctly), and it's infinitely better than having no formalized version numbering at all, but it's unsuitable for many upstreams' release processes.

One of the most notable things semver can't represent is non-linear releases: any time a project has a stable branch and a development branch, it's very easy for a release on the development branch to make it impossible to allocate a new version number for the stable branch, because the unique version number that semver would require it to have is already taken. For instance, suppose you have foo 1.2.3, and you fix a bug with changes that are complex and have a regression risk. The result has to be 1.2.4. Now try to fix a security vulnerability in 1.2.3, without pulling in the risky bug fixes; you can't, because semver says bug fixes always increment the micro version, but you already used version number 1.2.4.

The next-more-complex thing after semver is to restrict version numbers to be tuples of dot-separated integers (1.2.3.4 should be interpreted as (1, 2, 3, 4)), which are equally easy for something like RPM or dpkg to compare, and any reasonable version comparison function should sort them correctly. However, that doesn't allow for non-numeric version number parts like "1.5a" or "git20190108". Those are also fairly straightforward to add (obviously 1.5a should be less than 1.5b).

Introduce alphas, betas and release candidates and suddenly you also need RPM/dpkg semantics for `~`, which sorts less than anything, even the empty string (so that version 1.5 beta 1 can be represented as 1.5~beta1, which is less than 1.5); and now you already have 90% of the version comparison algorithm from RPM or dpkg, so you might as well have the whole thing.

PR #281, opened 2019-01-09
Link with -Bsymbolic if it's supported by the linker

kalev:
This should fix crashes when libappstream-glib and libappstream are both
linked into the same process by avoiding intra-library PLT jumps.

Flatpak master recently started linking with libappstream-glib, and this
caused issues in other apps that link with both libappstream and flatpak
(e.g. KDE Discover), now pulling in both libappstream-glib and
libappstream into the same process space.

Also bump required meson version to 0.46.0 for has_link_argument
support.

https://bugs.archlinux.org/task/61198

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -2,7 +2,7 @@ project('appstream-glib', 'c',
   version : '0.7.15',
   license : 'LGPL-2.1+',
   default_options : ['warning_level=1', 'c_std=c99'],
-  meson_version : '>=0.37.0'
+  meson_version : '>=0.46.0'
 )
 
 as_version = meson.project_version()
@@ -84,14 +84,25 @@ if get_option('dep11')
 endif
 
 # support stemming of search tokens
+cc = meson.get_compiler('c')
 if get_option('stemmer')
-  cc = meson.get_compiler('c')
   stemmer = cc.find_library('stemmer')
   conf.set('HAVE_LIBSTEMMER', 1)
 endif
 
 # use gperf for faster string -> enum matching
 gperf = find_program('gperf', required : true)
+
+global_link_args = []
+foreach arg: [ '-Wl,-Bsymbolic' ]
+  if cc.has_link_argument(arg)
+    global_link_args += arg
+  endif
+endforeach
+add_global_link_arguments(
+  global_link_args,
+  language: 'c'
+)
 
 gnome = import('gnome')
 i18n = import('i18n')


hughsie:
If CI passes, LGTM, thanks.

kalev:
@hughsie Thanks, CI should be fixed now.

@aleixpol Can you test this, please?

aleixpol:
Still got the freeze, now with these warnings:
```
(process:30227): GLib-GObject-WARNING **: 15:03:20.165: cannot register existing type 'AsIcon'

(process:30227): GLib-GObject-WARNING **: 15:03:20.165: cannot add private field to invalid (non-instantiatable) type '<invalid>'

(process:30227): GLib-CRITICAL **: 15:03:20.165: g_once_init_leave: assertion 'result != 0' failed

(process:30227): GLib-GObject-CRITICAL **: 15:03:20.165: g_object_new_with_properties: assertion 'G_TYPE_IS_OBJECT (object_type)' failed
```

Here's the backtrace:
```
#0 syscall() at /usr/lib/libc.so.6
#1 g_cond_wait() at /usr/lib/libglib-2.0.so.0
#2 g_once_init_enter() at /usr/lib/libglib-2.0.so.0
#3 as_icon_get_type() at /usr/src/debug/appstream-glib/libappstream-glib/as-icon.c:59
#4 AS_ICON() at /usr/src/debug/appstream-glib/libappstream-glib/as-icon.h:35
#5 as_icon_new() at /usr/src/debug/appstream-glib/libappstream-glib/as-icon.c:1032
#6 as_app_node_parse_child() at /usr/src/debug/appstream-glib/libappstream-glib/as-app.c:5032
#7 as_app_node_parse_full() at /usr/src/debug/appstream-glib/libappstream-glib/as-app.c:5443
#8 as_store_from_root() at /usr/src/debug/appstream-glib/libappstream-glib/as-store.c:1810
#9 as_store_from_xml() at /usr/src/debug/appstream-glib/libappstream-glib/as-store.c:2347
#10 ??() at /usr/lib/libflatpak.so.0
#11 ??() at /usr/lib/libflatpak.so.0
#12 ??() at /usr/lib/libflatpak.so.0
#13 ??() at /usr/lib/libflatpak.so.0
#14 flatpak_installation_get_installed_ref() at /usr/lib/libflatpak.so.0
#15 FlatpakBackend::getInstalledRefForApp() at /home/apol/devel/frameworks/discover/libdiscover/backends/FlatpakBackend/FlatpakBackend.cpp:223
#16 FlatpakBackend::updateAppState() at /home/apol/devel/frameworks/discover/libdiscover/backends/FlatpakBackend/FlatpakBackend.cpp:1070
#17 FlatpakBackend::addResource() at /home/apol/devel/frameworks/discover/libdiscover/backends/FlatpakBackend/FlatpakBackend.cpp:541
#18 FlatpakBackend::integrateRemote(_FlatpakInstallation*, _FlatpakRemote*)::$_25::operator()() const() at /home/apol/devel/frameworks/discover/libdiscover/backends/FlatpakBackend/FlatpakBackend.cpp:645
#19 QtPrivate::FunctorCall<QtPrivate::IndexesList<>, QtPrivate::List<>, void, FlatpakBackend::integrateRemote(_FlatpakInstallation*, _FlatpakRemote*)::$_25>::call(FlatpakBackend::integrateRemote(_FlatpakInstallation*, _FlatpakRemote*)::$_25&, void**)() at /usr/include/qt/QtCore/qobjectdefs_impl.h:146
#20 QtPrivate::Functor<FlatpakBackend::integrateRemote(_FlatpakInstallation*, _FlatpakRemote*)::$_25, 0>::call<QtPrivate::List<>, void>(FlatpakBackend::integrateRemote(_FlatpakInstallation*, _FlatpakRemote*)::$_25&, void*, void**)() at /usr/include/qt/QtCore/qobjectdefs_impl.h:256
#21 QtPrivate::QFunctorSlotObject<FlatpakBackend::integrateRemote(_FlatpakInstallation*, _FlatpakRemote*)::$_25, 0, QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*)() at /usr/include/qt/QtCore/qobjectdefs_impl.h:439
#22 QMetaObject::activate(QObject*, int, int, void**)() at /usr/lib/libQt5Core.so.5
#23 QFutureWatcherBase::event(QEvent*)() at /usr/lib/libQt5Core.so.5
#24 QApplicationPrivate::notify_helper(QObject*, QEvent*)() at /usr/lib/libQt5Widgets.so.5
#25 QApplication::notify(QObject*, QEvent*)() at /usr/lib/libQt5Widgets.so.5
#26 QCoreApplication::notifyInternal2(QObject*, QEvent*)() at /usr/lib/libQt5Core.so.5
#27 QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*)() at /usr/lib/libQt5Core.so.5
#28 ??() at /usr/lib/libQt5Core.so.5
#29 g_main_context_dispatch() at /usr/lib/libglib-2.0.so.0
#30 ??() at /usr/lib/libglib-2.0.so.0
#31 g_main_context_iteration() at /usr/lib/libglib-2.0.so.0
#32 QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>)() at /usr/lib/libQt5Core.so.5
#33 QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>)() at /usr/lib/libQt5Core.so.5
#34 QCoreApplication::exec()() at /usr/lib/libQt5Core.so.5
#35 main() at /home/apol/devel/frameworks/discover/discover/main.cpp:168
```

kalev:
Hm, maybe appstream needs a similar patch too.

ximion:
I don't see how Bsymbolic should help here, and "Unfortunately, –Bsymbolic is a dangerous option which can often result in some nonintuitive side effects." doesn't sound great either (after reading about that option).
The only options I see are:
  * Finally make just one AppStream library happen
  * Make Flatpak not use asglib / only use it internally
  * Use Valve/Collabora's libcapsule

The first option will take a few years I fear, the second option could work but I need to look at what Flatpak is actually using asglib for, and the last option is quite an odd hack as well.
I don't know of any way to not have the linker confused, and this error is not limited to just Discover but a lot of other software that uses libappstream and may also link against Flatpak.

smcv:
Normally, there would be a fourth option: use versioned symbols in both libappstream and appstream-glib, so that `as_whatever()` resolves to `as_whatever@LIBAPPSTREAM_0.x` or `as_whatever@APPSTREAM_GLIB_0.y`.

However, GObject APIs can't use that if they use the same namespacing prefix, because the GType registry is a flat global namespace.

smcv:
> Use Valve/Collabora's libcapsule

I don't think libcapsule helps you here. It's specifically for use with chroots and containers where a particular library might exist both "inside" and "outside".

If the Flatpak part and/or the part that uses libappstream are plugins, then you could use `dlmopen()`, which is the glibc function that libcapsule uses to get symbol namespaces; or maybe `RTLD_LOCAL` would even be enough.

smcv:
No, dlmopen doesn't help you either, because GObject type names are a flat global namespace, and both libappstream and appstream-glib want to use the name `AsIcon`.

smcv:
> Make Flatpak not use asglib / only use it internally

The only thing in the libflatpak.so.0 shared library that uses asglib is `flatpak_dir_load_appstream_store()`, which finds AppStream metadata for a remote and calls `as_store_from_file()`.

That seems to just be a convenience function used by `flatpak search` in the CLI, so it might be feasible to move it outside the shared library and into the flatpak(1) command-line tool. The FlatpakDir object should probably have a method to get the filename of the AppStream XML instead.

smcv:
> The only thing in the libflatpak.so.0 shared library that uses asglib is flatpak_dir_load_appstream_store()

Unfortunately that's out-of-date information, and appstream-glib code is now reachable from public APIs in the libflatpak shared library.

kalev:
@matthiasclasen fixed this on the flatpak side by dropping the libappstream-glib dependency from libflatpak and doing the appdata parsing manually using GMarkup. Let's close this PR here.

kalev:
https://github.com/flatpak/flatpak/pull/2580 is the fix.

ximion:
Nice! This resolves the issue for now, and fortunately the changes in Flatpak weren't actually that invasive or complicated (I feared Flatpak was using a lot more AppStream features via asglib).

ximion:
@hughsie and I are talking about that problem for a while now, and he recently suggested just to transition over to libappstream - the reason why asglib was originally created doesn't exist anymore for years. The only real issue is that it will take quite a bit of work and a lot of time to do that, if we don't want to risk breaking things.

At least there shouldn't be any behaviour differences between the libs anymore at the current point in time. I am working on finding and fixing them by spec clarifications or changes to asglib or libas, and Richard and I are talking about spec changes prior to making them to ensure there is no new divergence between asglib and libas. So in regards to raw AppStream spec compliance there shouldn't be an issue anymore, you only run into problem in cases like this (and obviously lost synergies when every feature has to be implemented twice).

PR #282, opened 2019-01-09
CI: Update Fedora from 25 to 29, and Debian from Stretch to Buster

kalev:
This gets us new meson for https://github.com/hughsie/appstream-glib/pull/281,
and also makes it possible to enable -Ddep11=true for Debian as it's no
longer blocked by too old meson in Stretch.


hughsie:
Assuming CI passes, LGTM.

PR #284, opened 2019-01-22
Add support for per-release URLs

hughsie:


libappstream-glib/as-release.c
```diff
@@ -1019,6 +1097,17 @@ as_release_node_parse_dep11 (AsRelease *release, GNode *node,
 			}
 			continue;
 		}
+		if (g_strcmp0 (tmp, "Url") == 0) {
```
ximion: I'm quite sure that's a lower-cased `url` - in the YAML, only the top nodes are upper-cased, every child node is lower-cased all the way.

hughsie: God catch, thanks.

libappstream-glib/as-release.c
```diff
@@ -769,6 +774,67 @@ as_release_set_description (AsRelease *release,
...
+ * @release: a #AsRelease instance.
+ *
+ * Gets the URLs set for the application.
+ *
+ * Returns: (transfer none): hash table of URLs
+ *
+ * Since: 0.7.15
+ **/
+GHashTable *
+as_release_get_urls (AsRelease *release)
```
kalev: Is this actually needed as public API? Seems to leak a lot of internal details (hash table).

libappstream-glib/as-release.c
```diff
@@ -769,6 +774,67 @@ as_release_set_description (AsRelease *release,
...
+ * @url_kind: the URL kind, e.g. %AS_URL_KIND_HOMEPAGE.
+ *
+ * Gets a URL.
+ *
+ * Returns: string, or %NULL if unset
+ *
+ * Since: 0.7.15
+ **/
+const gchar *
+as_release_get_url_item (AsRelease *release, AsUrlKind url_kind)
```
kalev: Maybe call it just as_release_get_url() ? _item at the end seems unnecessary and just makes it more of a mouthful.

libappstream-glib/as-release.c
hughsie: I just coped the design (mistake?) from AsApp. I can nuke this, I'm not using it at the moment. Ideas welcome.

libappstream-glib/as-release.c
hughsie: I just coped the name (another mistake?) from AsApp. Maybe `as_release_get_url_by_kind`?

libappstream-glib/as-release.c
kalev: I think I'd leave it out then and add it only when something actually needs it.

libappstream-glib/as-release.c
kalev: Sure, works for me!

ximion: libappstream just calls this `as_release_get_url(rel, kind)` and has a `as_release_get_url(rel, kind, url)`

libappstream-glib/as-enums.h
```diff
@@ -84,6 +85,7 @@ typedef enum {
 	AS_URL_KIND_HELP,		/* Since: 0.1.5 */
 	AS_URL_KIND_MISSING,		/* Since: 0.2.2 */
 	AS_URL_KIND_TRANSLATE,		/* Since: 0.6.1 */
+	AS_URL_KIND_DETAILS,		/* Since: 0.7.15 */
```
ximion: In libappstream I created a `AS_RELEASE_URL_KIND` to distinguish between what URLs are valid in releases and which ones aren't.
That distinction is probably not necessary though.

libappstream-glib/as-release.c
kalev: I think as_release_get_url(rel, kind) would be my preference as well, but as_release_get_url_by_kind() works as well.

hughsie: as_release_get_url it is then. thanks.

libappstream-glib/as-release.c
hughsie: Okay, agree. Thanks.

kalev:
Looks good to me!

Issue #252, opened 2018-07-20
Windows support ?


Salamandar:
Hi,
appstream-glib currently relies on several linux-only components:

* `dependency('uuid')` that is provided by libutil-linux/libutil-linux-dev on msys… And unavailable on mingw.
* `dependency('gio-unix')`
* header `fnmatch.h`

Jehan:
Hi! I'd like to back a bit this request, since what Salamandar didn't specify, this is that we need this for GIMP. I don't really think AppStream metadata really needs to be platform specific. :-)

I tried to cross-build util-linux (as well as this other implementation which is apparently independent, whose website seems down though it can still be found on [Fedora repos](https://src.fedoraproject.org/repo/pkgs/rpms/uuid/uuid-1.6.2.tar.gz/5db0d43a9022a6ebbbc25337ae28942f/) and [archive.org](https://web.archive.org/web/20171228094631/http://www.ossp.org/pkg/lib/uuid)) but failed so far. :-/

ximion:
Just out of curiosity, what does GIMP actually need appstream-glib for?
Does it actually have to read AppStream metadata somewhere, and if so, why?

Jehan:
> Just out of curiosity, what does GIMP actually need appstream-glib for?

I am writing an extension management system (searching, (un)installing, updating extensions from within GIMP) for plug-ins, brushes and whatever data. The base of this is the metadata which a plug-in creator will write to describe the extension and its contents.

I could have used another format or even make or own metadata format, but thought it was smarter to just use AppStream since it has everything we need.

ximion:
@Jehan That's pretty cool!
As far as I can see, you will have a few options to achieve that:
* Use appstream-glib: You should definitely disable the builder part when compiling on Windows, that should get rid of a great portion of dependencies. To cut down the rest though, asglib would need changes and I am not sure if those could be done without an API break. @hughsie knows more.
* Use `appstream:` links: Just use the AppStream URL handler to launch the software center to install a missing dependency. That is likely the easiest way possible - at least on Linux - to install missing components, and means you don't need to implement any complext installation logic yourself. It's also easy to use for the user. See https://www.freedesktop.org/software/appstream/docs/sect-AppStream-Services-UrlHandler.html for details.
[...]

PR #288, opened 2019-02-06
Wip/jehan/win32 port

Jehan:
As discussed on #252, this ports appstream-glib to Windows.

With all the changes done, I could successfully build the library, it works on GIMP, and the unit tests work fine too.

Though note that I configured with many options false:

```
Project options:
  Option        Current Value Possible Values Description                        
  ------        ------------- --------------- -----------                        
  alpm          false         [true, false]   enable ALPM support                
  builder       false         [true, false]   enable AppStream builder           
  dep11         false         [true, false]   enable DEP-11                      
  fonts         true          [true, false]   enable font support                
  gtk-doc       false         [true, false]   generate API reference             
  introspection false         [true, false]   generate GObject Introspection data
  man           true          [true, false]   generate man pages                 
  rpm           false         [true, false]   enable RPM support                 
  stemmer       false         [true, false]   enable stemmer support 
```

For the time being, I don't think I need most of these (for GIMP extensions, for which we want to add management through metadata; as a reminder of why I need these). If I realize I do, I would gladly try to port whatever part of code these options need. But for now, I went the easy route (which was annoying enough as it is! :P).

### DIFF
--- a/client/as-util.c
+++ b/client/as-util.c
@@ -23,8 +23,11 @@
 
 #include <glib/gi18n.h>
 #include <glib/gstdio.h>
-#include <glib-unix.h>
 #include <gio/gio.h>
+
+#ifndef _WIN32
+#include <glib-unix.h>
+#endif
 
 #include <appstream-glib.h>
 #include <archive_entry.h>
@@ -4282,6 +4285,20 @@ as_util_watch_cancelled_cb (GCancellable *cancellable, gpointer user_data)
 	g_main_loop_quit (priv->loop);
 }
 
+#ifdef _WIN32
+static AsUtilPrivate *as_util_priv = NULL;
+
+static BOOL
+as_util_sigint_cb (DWORD dwCtrlType)
+{
+    if (dwCtrlType == CTRL_C_EVENT) {
+        g_debug ("Handling SIGINT");
+        g_cancellable_cancel (as_util_priv->cancellable);
+        return TRUE;
+    }
+	return FALSE;
+}
+#else
 static gboolean
 as_util_sigint_cb (gpointer user_data)
 {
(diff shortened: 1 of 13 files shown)


hughsie:
@Jehan looks great! I've simplified it down in my branch -- removing some of the stuff you're not going to be using. Don't worry tho, I know the API :) Can you have a look at wip/hughsie/win32 and see if it work for you please?

Jehan:
Argh you removed all my hard work on using Win32 Uuid API, fnmatch equivalent and whatnot! 😛

Well if you are sure we won't need these in GIMP (where we will mostly need the lib to read appdata per extension, as well as extension store appdata for installation), I guess it's ok. But for the record, for stuff like all the GUID functions for instance, I checked it worked the same (the unit tests on GUID at the very least worked fine with all expected results).
Yet if you prefer to just set these as "unsupported" on Windows and are sure we won't need these anyway (now or in the future; I have not looked in details how/why it is used), I'm not going to fight. :-)

> see if it work for you please?

Not going to test today, as I have some other priorities. But will crossbuild your branch and test GIMP in a Windows VM early next week. Thanks for the review! :-)

Jehan:
@hughsie Well in the end, I lied to myself that I would not be intrigued enough to want and test today! So I just tried to cross-build your branch for Win and it failed with a few errors. Did you manage to build for Windows?

Should I make patches for this branch?

Jehan:
Ok I just created #289 to fix the build of wip/hughsie/win32 (just 3 easy errors). :-)

With these few fixes, it builds fine, and my GIMP master on a Windows 10 VM runs just fine. Appdata files are properly parsed. All good so far.

Still a bit sad that the stuff I spent the most time on aren't included, but well… the main purpose is to use appdata files for GIMP extensions. And if you say that these other piece of code will definitely not be useful, then I won't be sick over it! :P

hughsie:
> Still a bit sad that the stuff I spent the most time on aren't included

Sorry; I didn't want to include complicated stuff that wasn't going to be used. I might end up using the GUID stuff on fwupd in the future, so I'll acknowledge you when that happens :)

hughsie:
Okay, tracked in https://github.com/hughsie/appstream-glib/pull/290

PR #290, opened 2019-02-08
Allow building on Windows

hughsie:
Signed-off-by: Richard Hughes <richard@hughsie.com>

### DIFF
--- a/client/as-util.c
+++ b/client/as-util.c
@@ -23,8 +23,11 @@
 
 #include <glib/gi18n.h>
 #include <glib/gstdio.h>
-#include <glib-unix.h>
 #include <gio/gio.h>
+
+#ifndef _WIN32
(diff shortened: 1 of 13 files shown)


Jehan:
Any reasons why this is not being merged? :-)

We will need appstream to be Windows-compatible before being able to release GIMP with our extension system. And since releases are always a bit delayed, better too soon that late!
 Actually we'd probably need the lib to be macOS compatible too, but I don't have the possibility to do this port (unless it already works there, no idea).

hughsie:
No reason at all, other than overworked and underpaid :)

Jehan:
Ahah! Thanks @hughsie!

I don't mean to participate to over-exerting you! So have some rest too and thanks for the awesome work! ;-)

PR #292, opened 2019-02-10
Fix build of libappstream-glib clients

juergbi:
With '#pragma once', __APPSTREAM_GLIB_H is never defined.

        In file included from /usr/include/libappstream-glib/appstream-glib.h:35:
        /usr/include/libappstream-glib/as-utils.h:11:2: error: #error "Only <appstream-glib.h> can be included directly."
         #error "Only <appstream-glib.h> can be included directly."
          ^~~~~

Fixes: 87b7483e ("Use '#pragma once' to avoid a lot of boilerplate")

### DIFF
--- a/libappstream-glib/as-agreement-section.h
+++ b/libappstream-glib/as-agreement-section.h
@@ -7,7 +7,7 @@
 
 #pragma once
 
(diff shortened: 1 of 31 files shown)


juergbi:
This currently breaks gnome-build-meta: https://gitlab.gnome.org/GNOME/gnome-build-meta/-/jobs/210604

Issue #263, opened 2018-09-12
Add an option that checks for Flathub requirements


hadess:
Not sure what to call it, but having appstream-compose fail with errors when an appdata file wouldn't be accepted by Flathub would be of interest:
- Check for launchable desktop-id presence
- Check for .desktop file validity (see https://github.com/hughsie/appstream-glib/pull/262 and https://github.com/flathub/org.freedesktop.Piper/issues/1#issuecomment-420446582)

etc.

barthalion:
From things we "broke" often in past:

* Ensure `<description>` content is wrapped in `<p>`
* Ensure OARS section is present
* A warning if releases section is missing

hughsie:
I'm hesitant to encode the flathub requirements into libappstrea-glib -- people are going to be preparing the flathub builds on their distro, and that's certainly not going to have the most up to date `appstream-util validate` -- and even if they're running `appstream-compose` from the sandbox the appstream-glib in the runtime they're using might be quite old still.

This really feels like it should be part of flatpack-builder, or even better as part of the flathub self tests that get run when you build the component for the first time.

hadess:
> I'm hesitant to encode the flathub requirements into libappstrea-glib -- people are going to be preparing the flathub builds on their distro, and that's certainly not going to have the most up to date `appstream-util validate` -- and even if they're running `appstream-compose` from the sandbox the appstream-glib in the runtime they're using might be quite old still.

Those requirements are mostly the same ones as for making an app show up in GNOME Software.

> This really feels like it should be part of flatpack-builder, or even better as part of the flathub self tests that get run when you build the component for the first time.

Flatpak-builder doesn't parse the appdata files, but uses appstream-compose to merge metadata into the appdata files it will eventually serve.

I don't mind not having a "--flathub" option, but that would mean adding separate checks for each one of the common bugs above. I don't mind that, it would just mean more work. Given how much time I've spent on trying to fix a  one-character bug in the appdata file, I'd rather the tools warned me early rather than letting me send broken appdata files to flathub.

ramcq:
> even if they're running appstream-compose from the sandbox the appstream-glib in the runtime they're using might be quite old still.

[...]

PR #296, opened 2019-03-20
Modernize the validation requirements

hughsie:
* Require <content_rating> for any desktop or console application
 * Make <release> mandatory when validating
 * Further relax style requirements
 * Require description for desktop an console apps

This changes a lot of things that are now best practice for all apps, and
required to be included in Flathub.

Use `appstream-util validate-relax` to ignore some new failures and use
`appstream-util validate-strict` to bring them back.

Fixes: https://github.com/hughsie/appstream-glib/issues/263

### DIFF
--- a/libappstream-glib/as-app-validate.c
+++ b/libappstream-glib/as-app-validate.c
@@ -147,10 +147,16 @@ as_app_validate_has_first_word_capital (AsAppValidateHelper *helper, const gchar
 static void
 as_app_validate_description_li (const gchar *text, AsAppValidateHelper *helper)
 {
-	gboolean require_sentence_case = TRUE;
+	gboolean require_sentence_case = FALSE;
 	guint str_len;
-	guint length_li_max = 100;
-	guint length_li_min = 20;
+	guint length_li_max = 500;
+	guint length_li_min = 3;
+
+	/* make the requirements more strict */
+	if ((helper->flags & AS_APP_VALIDATE_FLAG_STRICT) > 0) {
+		require_sentence_case = TRUE;
+		length_li_max = 250;
+	}
 
 	/* relax the requirements a bit */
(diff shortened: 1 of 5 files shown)


libappstream-glib/as-app-validate.c
```diff
@@ -1299,6 +1353,42 @@ as_app_validate (AsApp *app, guint32 flags, GError **error)
...
+			const gchar *cat = g_ptr_array_index (categories, i);
+			for (guint j = 0; cats[j] != NULL; j++) {
+				if (g_strcmp0 (cats[j], cat) == 0)
+					nr_toplevel_cats++;
+			}
+		}
+		if (nr_toplevel_cats == 0) {
+			ai_app_validate_add (helper,
+					     AS_PROBLEM_KIND_TAG_MISSING,
+					     "main <category> not present");
```
kalev: I wonder if `<category> must include one of the Main Categories listed in the desktop entry spec` would be a bit clearer?

libappstream-glib/as-self-test.c
```diff
@@ -2448,7 +2436,7 @@ as_test_app_validate_style_func (void)
 	_as_app_add_format_kind (app, AS_FORMAT_KIND_APPDATA);
 	as_app_set_metadata_license (app, "BSD");
 	as_app_set_project_license (app, "GPL-2.0+");
-	as_app_set_name (app, "C", "Test app name that is very log indeed.");
+	as_app_set_name (app, "C", "Test app name that is very log indeed although Rob forced me to relax the requirements.");
```
kalev: log -> long

libappstream-glib/as-app-validate.c
hughsie: Agreed, thanks.

Issue #302, opened 2019-05-13
Stricter validation hasn't been adjusted in the documentation


kparal:
My application on Flathub started [failing to get built](https://flathub.org/builds/#/builders/32/builds/3285), probably due to #296:
```
$ flatpak run org.freedesktop.appstream-glib validate builddir/*/share/appdata/org.freefilesync.FreeFileSync.appdata.xml
builddir/files/share/appdata/org.freefilesync.FreeFileSync.appdata.xml: FAILED:
• tag-missing           : <translation> not specified
• tag-missing           : <update_contact> is not present
• attribute-invalid     : <screenshot> width too small [https://www.freefilesync.org/images/screenshots/RealTimeSync.png] minimum is 624px
Validation of files failed
program finished with exit code 1
```
However, when I look into the documentation:
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html
I don't any violations. One by one:
```
• tag-missing           : <translation> not specified
```
> The <translation/> tag is an **optional** tag which...

```
• tag-missing           : <update_contact> is not present
```
> The <update_contact/> tag is an **optional** tag which...

```
• attribute-invalid     : <screenshot> width too small [https://www.freefilesync.org/images/screenshots/RealTimeSync.png] minimum is 624px
```
> **Ideally**, all screenshots should have a 16:9 aspect ratio, and should have a width that is no smaller than 620 pixels.

There is no minimum size requirement.

My appdata file seems to conform to the specification. I also checked appstream-glib readme file (I haven't found any extra documentation for it) and I didn't find any extra requirements on top of the specification.

I want to request the documentation to be updated. Or is there a reason for the spec and the implementation to not be in sync?

Furthermore, I'd like to request clearer explanation in the docs. I'm quite confused by the `<translation> not specified` error. Looking at the documentation, how is a packager who is not a programmer supposed to find out whether the translation system used is gettext or qt? What about proprietary apps using a different translation system? What is a translation domain (is it some gettext jargon?) and who am I supposed to find it out? And why exactly is this mandatory?

Similarly, the error and docs about `<update_contact/>` doesn't seem to address that some projects might not have a public email contact. Or am I supposed to put my address in, as a packager? The spec is confusing in here. Again, does it have to be mandatory?

Thanks.

hughsie:
> My appdata file seems to conform to the specification.

[...]

PR #303, opened 2019-05-14
Do not require <update_contact> for validation unless strict

hughsie:
Fixes some of https://github.com/hughsie/appstream-glib/issues/302


kalev:
+1 from me!

PR #310, opened 2019-07-16
Do not trigger an inotity event when the AppStream XML data is unchanged

hughsie:
Do not modify the ctime or mtime in as_utils_install_xml() if the SHA1
checksums are identical. This means gnome-software starts without triggering an
XbSilo rebuild if we refresh the repos with PackageKit as dnf is so keen to do.

### DIFF
--- a/libappstream-glib/as-utils.c
+++ b/libappstream-glib/as-utils.c
@@ -1143,8 +1143,12 @@ as_utils_install_xml (const gchar *filename,
 {
 	gchar *tmp;
 	g_autofree gchar *basename = NULL;
+	g_autofree gchar *checksum_dest = NULL;
+	g_autofree gchar *checksum_src = NULL;
 	g_autofree gchar *path_dest = NULL;
 	g_autofree gchar *path_parent = NULL;
+	g_autoptr(GBytes) blob_dest = NULL;
+	g_autoptr(GBytes) blob_src = NULL;
 	g_autoptr(GFile) file_dest = NULL;
 	g_autoptr(GFile) file_src = NULL;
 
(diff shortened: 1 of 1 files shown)


hughsie:
> I don't understand what this is doing. 

Right, I agree it's not simple. With AppStream metadata coming from distros, COPRs and Flathub we can't trust the `<components origin="foo">` either actually exists or is the right value. On opensuse, I believe all the metadata has the same origin, and then when PackageKit calls install-origin we set the origin stored in the file as the repo name. This is how we know where to look for the icons, and also which repo provides the app. I agree it's inelegant and decompressing, parsing, exporting and compressing each time just to change a simple string is somewhat insane.

> Why do you need to recompress the data and how is it related to not changing mtime?

At the moment the existing code copies the source to the destination, loads the source, sets the origin then saves it to the destination, overwriting the last written copy. This means when src=dest we actually get 2 inotify signals about the file, which marks the silo as invalid.

> Maybe just avoid copying if the checksum is identical? 

Checking the mtime is "free", as in it's just one "stat" call. Doing a checksum for all the multi-megabyte metadata files (and all the hundreds of tiny appdata/desktop files all over the disk) would be hugely expensive which would kill the start-up time gains we got when switching to the zero-copy xmlb design.

I guess the other thing we could do is find the root problem; like finding out why PackageKit/libdnf is copying over the file (and hopefully not downloading) when we've explicitly asked for a cache timeout of 0.

kalev:
Ahh, thanks for the explanation. It does seem like a wrong layer for fixing the issue, but if you think it's the easiest way for now, works for me :)

hughsie:
No, you're right to tell me it's insane. I think it's a PK mistake anyway.

hughsie:
I think the real bug is https://github.com/hughsie/PackageKit/pull/330 -- thanks for the (in)sanity check.

PR #311, opened 2019-07-22
Support loading YAML from as_store_from_bytes

robert-ancell:
Added to resolve [GNOME Software not loading YAML](https://gitlab.gnome.org/GNOME/gnome-software/issues/742).

### DIFF
--- a/libappstream-glib/as-store.c
+++ b/libappstream-glib/as-store.c
@@ -1830,30 +1830,32 @@ as_store_from_root (AsStore *store,
 }
 
 static gboolean
-as_store_load_yaml_file (AsStore *store,
-			 GFile *file,
-			 AsAppScope scope,
-			 GCancellable *cancellable,
-			 GError **error)
+is_dep11_data (GBytes *bytes)
+{
+	const gchar *data;
+	gsize size;
+
+	/* look for DEP-11 header */
+	data = g_bytes_get_data (bytes, &size);
+	return g_strstr_len (data, size, "File: DEP-11") != NULL;
+}
+
+static gboolean
+load_yaml (AsStore *store,
+	   AsYaml *root,
+	   const gchar *source_filename,
+	   AsAppScope scope,
+	   GCancellable *cancellable,
+	   GError **error)
 {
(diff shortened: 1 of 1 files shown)


robert-ancell:
The DEP-11 check is a bit hacky, wasn't sure if that should be in as-yaml.c.

libappstream-glib/as-store.c
```diff
@@ -2269,6 +2326,10 @@ as_store_from_bytes (AsStore *store,
 		return as_store_from_xml (store, tmp, NULL, error);
 	}
 
+	/* is a DEP-11 file */
+	if (g_strcmp0 (content_type, "text/plain") == 0 && is_dep11_data (bytes))
```
hughsie: Any idea why yaml files are detected as `text/plain`?

robert-ancell: I don't think it's possible to reliably detect YAML as there's nothing consistently used in the start of a YAML file that would reliably indicate that. Even the `is_dep11_data` function is pretty vague in checking, though it shouldn't collide with any other appstream file format and will fail if it turns out not to be YAML.

libappstream-glib/as-store.c
```diff
@@ -1830,31 +1830,46 @@ as_store_from_root (AsStore *store,
...
+	data = g_bytes_get_data (bytes, &size);
+	line_start = 0;
+	for (i = 0; i < 1000 && i < size; i++) {
+		gsize line_length;
+		if (data[i] != '\n')
+			continue;
+
+		/* look for DEP-11 header */
+		line_length = i - line_start;
+		if (strncmp (data + line_start, "File: DEP-11", line_length) == 0)
```
hughsie: this seems a bit clumsy -- can't we just do something like `g_strstr_len (data, "File: DEP-11", size) != NULL`

robert-ancell: I thought that would be not a particularly strong indication, but if it's not a DEP-11 file then it's going to fail in `as_store_load_yaml_data` so that seems reasonable.

PR #316, opened 2019-08-17
its: Sync with appstream

piotrdrag:
Integrates these commits:

https://github.com/ximion/appstream/commit/e42dd0d071f27d74419b04a52a8388e00777cd88
https://github.com/ximion/appstream/commit/71d134e82752be6f8851c65ba0ce8c312a72214c
https://github.com/ximion/appstream/commit/8c939f999ebf82c3bbf5ad179b24fd7478a0f1d0
https://github.com/ximion/appstream/commit/51ab4f32c56f5cdc293ba0512f003542b83713fc

### DIFF
--- a/data/appdata.its
+++ b/data/appdata.its
@@ -1,14 +1,28 @@
 <?xml version="1.0"?>
 <its:rules xmlns:its="http://www.w3.org/2005/11/its"
            version="2.0">
+
+  <its:withinTextRule withinText="yes" selector="/component//description//em   |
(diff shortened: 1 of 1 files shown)


hughsie:
I think this is exactly what we want, but I think we need to add support for the new AppStream elements before merging these patches.

piotrdrag:
I added another recent appstream commit.

piotrdrag:
Added one more.

PR #321, opened 2019-09-13
Log expected and actual screenshot dimensions when failing

muelli:
Sometimes it is quite helpful to see the values the validator objects.
Especially when it is hard to comprehend why the validation fails if one has,
say, an image in a stale cache which the validator does not have.


hughsie:
In concept, yes, but it needs to compile :)

muelli:
whoops. silly me. fortunately, there is CI :)

Issue #323, opened 2019-09-18
appstream-util doesn't accept MIT license as metadata license


scx:
`appstream-util` doesn't accept the `MIT` license as a metadata license.

According to the Freedesktop specification, `MIT` is a valid license (and even recommended!).
https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#qsr-app-contents
https://spdx.org/licenses/

Unfortunately, `appstream-util`/`org.freedesktop.appstream-glib` doesn't accept it anymore, at least as a metadata license:
```
$ flatpak run org.freedesktop.appstream-glib validate build/files/share/appdata/com.notepadqq.Notepadqq.appdata.xml
build/files/share/appdata/com.notepadqq.Notepadqq.appdata.xml: FAILED:
• tag-invalid           : <metadata_license> is not valid [MIT]
```

```
$ appstream-util validate build/files/share/appdata/com.notepadqq.Notepadqq.appdata.xml
build/files/share/appdata/com.notepadqq.Notepadqq.appdata.xml: FAILED:
• tag-invalid           : <metadata_license> is not valid [MIT]
• tag-missing           : <translation> not specified
• tag-missing           : <update_contact> is not present
Validation of files failed
```

However, it is still OK for `validate-relax`:
```
$ flatpak run org.freedesktop.appstream-glib validate-relax build/files/share/appdata/com.notepadqq.Notepadqq.appdata.xml
build/files/share/appdata/com.notepadqq.Notepadqq.appdata.xml: OK
```
```
$ appstream-util validate-relax build/files/share/appdata/com.notepadqq.Notepadqq.appdata.xml
build/files/share/appdata/com.notepadqq.Notepadqq.appdata.xml: OK
```

And of course, `appstreamcli` has no problem with that:
```
$ appstreamcli validate build/files/share/appdata/com.notepadqq.Notepadqq.appdata.xml
Validation was successful: pedantic: 1
```

Temporary patch ([0007-Update-list-of-allowed-metadata-licences.patch](https://pastebin.com/raw/21eKT5vH)):
```
diff --git a/libappstream-glib/as-app-validate.c b/libappstream-glib/as-app-validate.c
index ccac4e2..c1103ac 100644
--- a/libappstream-glib/as-app-validate.c
+++ b/libappstream-glib/as-app-validate.c
@@ -1046,6 +1046,12 @@ as_app_validate_license (const gchar *license_text, GError **error)
 static gboolean
[...]

PR #324, opened 2019-09-18
Update list of allowed metadata licences

scx:
This basically updates the list with values from the reference implementation.
https://github.com/hughsie/appstream-glib/blob/d9270c52bffde6ebd4fc3e3b4e0d34f802b920db/libappstream-glib/as-app-validate.c#L1046-L1068
https://github.com/ximion/appstream/blob/08f44a7fda48dc639346414687e79ee2e82963dd/src/as-spdx.c#L469-L507

Fixes hughsie/appstream-glib#323

### DIFF
--- a/libappstream-glib/as-app-validate.c
+++ b/libappstream-glib/as-app-validate.c
@@ -1046,6 +1046,12 @@ as_app_validate_license (const gchar *license_text, GError **error)
 static gboolean
 as_app_validate_is_content_license_id (const gchar *license_id)
 {
+	if (g_strcmp0 (license_id, "@FSFAP") == 0)
+		return TRUE;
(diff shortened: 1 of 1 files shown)


hughsie:
I've been told I can merge this. The legal team did say the list of permissible licences "was getting ridiculous" so please @ximion please don't add any more upstream. I have to do a whole heap of "paperwork" now to change the superset licence of appstream-data in Fedora and RHEL now.

ximion:
@hughsie Those have been permitted for a really long time and I don't intend to add more.
In fact, I should probably change the specification from saying "any permissive license is fine, but AppStream implementations may restrict the list of permitted licenses" to "this list of permissive licenses is allowed explicitly, and nothing else". Now that both implementations support the same stuff, that's probably very sensible to do.

PR #325, opened 2019-09-27
Allow parsing desktop files using as_app_parse_data()

hughsie:
This fixes showing local apps in GNOME Software without AppStream information.

### DIFF
--- a/libappstream-glib/as-app-desktop.c
+++ b/libappstream-glib/as-app-desktop.c
@@ -456,31 +456,9 @@ as_app_parse_file_key_fallback_comment (AsApp *app,
 }
 
 gboolean
-as_app_parse_desktop_file (AsApp *app,
-			   const gchar *desktop_file,
-			   AsAppParseFlags flags,
-			   GError **error)
+as_app_parse_desktop_kf (AsApp *app, GKeyFile *kf, AsAppParseFlags flags, GError **error)
 {
-	GKeyFileFlags kf_flags = G_KEY_FILE_KEEP_TRANSLATIONS;
-	gchar *tmp;
-	guint i;
-	g_autoptr(GError) error_local = NULL;
-	g_autofree gchar *app_id = NULL;
-	g_autoptr(GKeyFile) kf = NULL;
 	g_auto(GStrv) keys = NULL;
-
-	/* load file */
-	kf = g_key_file_new ();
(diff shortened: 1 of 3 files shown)


uajain:
Sure, LGTM

kalev:
Seems fine to me. This just breaks out a function so you can call it in `as_app_parse_data` for desktop files, right?

hughsie:
> This just breaks out a function so you can call it in as_app_parse_data for desktop files, right

Exactly that.

libappstream-glib/as-app-desktop.c
```diff
@@ -456,32 +456,10 @@ as_app_parse_file_key_fallback_comment (AsApp *app,
 }
 
 gboolean
-as_app_parse_desktop_file (AsApp *app,
-			   const gchar *desktop_file,
-			   AsAppParseFlags flags,
-			   GError **error)
+as_app_parse_desktop_kf (AsApp *app, GKeyFile *kf, AsAppParseFlags flags, GError **error)
```
uajain: @hughsie This function wasn't supposed to be a public api right? 
missing `static`

uajain: Pardon me, but I don't see any call to `as_app_parse_desktop_kf` in [as-app.c](https://github.com/hughsie/appstream-glib/blob/master/libappstream-glib/as-app.c)
Am I missing something?

hughsie: My mistake, apologies! I thought this was about as_app_parse_desktop_data() -- i'll staicize the method now.

hughsie: Fixed in 1729786131ac68ec3f8dd100af484d23cb85ce65

uajain: > Fixed in 1729786131ac68ec3f8dd100af484d23cb85ce65

Thanks. 

But I found yet another problem.
The symbols for `as-app-private.h` are actually getting exported. See in the [debian packaging](https://salsa.debian.org/pkgutopia-team/appstream-glib/blob/debian/0.7.16-1/debian/libappstream-glib8.symbols).

Asking around, I was advised that `G_GNUC_INTERNAL` should prevent exporting private-api as symbols. Not super-familiar with this stuff so can you please take a look what is going wrong? 
Doesn't RPM packaging also get all these private headers as symbols? :thinking:

PR #335, opened 2019-11-14
Revert "Allow parsing desktop files using as_app_parse_data()"

kalev:
The original issue was fixed in gnome-software in a different way that
didn't require appstream-glib changes in
https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/335.

This commit reverts the appstream-glib change as it regressed older
gnome-software versions.

This reverts commit 3b6701a62ef1e8f449010bd1c2bba01d1929b560.
This reverts commit d407862968c6528f44c93102e53da5664ffc8f4d.

### DIFF
--- a/libappstream-glib/as-app-desktop.c
+++ b/libappstream-glib/as-app-desktop.c
@@ -455,78 +455,6 @@ as_app_parse_file_key_fallback_comment (AsApp *app,
 	return TRUE;
 }
 
-static gboolean
-as_app_parse_desktop_kf (AsApp *app, GKeyFile *kf, AsAppParseFlags flags, GError **error)
-{
-	g_auto(GStrv) keys = NULL;
-
-	/* check this is a valid desktop file */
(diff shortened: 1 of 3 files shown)


kalev:
This would allow updating libappstream-glib in F30. @hughsie Am I correct that this should no longer be needed?

hughsie:
@kalev I don't think we can do this, gnome-software is actually using as_app_parse_data() to avoid spewing warnings everywhere. I think we can backport the GS 314cde9ccff4f0a9eb685b174f7add0f31b1e9c7 commit into f30 tho.

kalev:
Ah, sure, works for me. Thanks!

kalev:
I went ahead and did that in https://src.fedoraproject.org/rpms/gnome-software/c/22f0b24c5013c6632ff191eb7d3cf8a0e4ba0dba?branch=f30

PR #336, opened 2019-11-21
Fix Win32 headers.

Jehan:
For some reason, with mingw-w64 64-bit builds, this was just a warning.
But building for 32-bit, linking was failing:

> libappstream-glib/as-self-test.c:99: undefined reference to `GetFullPathNameA'
> collect2: error: ld returned 1 exit status

Adding the appropriate header fixes both the warning and error.


hughsie:
Also, I wonder if `_fullpath` would be a better replacement...

Jehan:
> Doesn't this include windef.h for us?

Checking `windows.h` contents (the ones provided by the mingw-w64 project), it indeed includes `windef.h`. And re-building both 32 and 64-bit, it builds fine. So let's say it's good. I amended the commit.

> Also, I wonder if _fullpath would be a better replacement...

I didn't understand the comment.

hughsie:
> I didn't understand the comment.

This is what i came up with in fwupd:  https://github.com/fwupd/fwupd/pull/1577/commits/2afea05dc51da34eb90ee79ec2deffe9ab6456f0#diff-9f39078dee82f226975a6beb2b04d62cR1499

It very well may be wrong, if so please say! :)

Jehan:
Ah ok, I didn't think it was an actual function name (because, what a weird idea to have a public API starting with underscore and different name style!).

> It very well may be wrong, if so please say! :)

Well now that I know and looked at the docs, both function look quite similar. So I tested, and using `_fullpath()` builds correctly (I didn't run the code, only cross-built it). No error, no warning.

So if you want, I can commit and push this change.

Other than this, I can't say if it is better or whatever.

Jehan:
Well actually looking closer at Win32 docs, it says that `GetFullPathNameA()` is not recommended for multi-threaded applications, and since `appstream-glib` is a library, it can definitely be used in multi-threaded cases. So whatever. I made the change.

Jehan:
> I still think you need to test for EXISTS as it returns non null if

Indeed none of these functions test for existence. Yet you already test for existence (just before using relative paths), but only from `TESTDIRSRC`. I can also add another `g_file_test()` from `TESTDIRBUILD` and if it fails, return `NULL` directly instead of continuing. Is that what you are asking?

hughsie:
I've pushed a slightly modified version of this, thanks!

PR #338, opened 2019-12-03
as-app: properly initialize unique_id_mutex

Keruspe:
We are using it, clearing it, but e never initialize it.
For some reasons it seems to work, not sure how or why,
but when glib is uilt with clang, then the problem
surfaces and we get a nice segmentation fault.
Properly initializing it fixes this


hughsie:
Great, thanks!

PR #346, opened 2020-01-17
Fix build on the macOS platform

samm-git:
This PR includes:

- use darwin native uuid library instead of external one
- support version script with clang/apple
- use `.so` suffix for internal plugins


libappstream-glib/appstream-glib.map.clang
```diff
@@ -0,0 +1 @@
+_as_*
```
hughsie: How come the leading underscrore?

samm-git: this is standard apple/clang syntax, all symbols are exported with `_` prefix to the target library and this is the way linker searching them. I checked result with `nm` took already

samm-git: JFYI, resulted symbols are looking that way:

```
bash-3.2$ nm -gU libappstream-glib.8.dylib |head
0000000000023b30 T _as_agreement_add_section
0000000000023750 T _as_agreement_get_kind
0000000000023a80 T _as_agreement_get_section_default
00000000000239e0 T _as_agreement_get_sections
0000000000023610 T _as_agreement_get_type
0000000000023890 T _as_agreement_get_version_id
00000000000236e0 T _as_agreement_kind_from_string
00000000000236b0 T _as_agreement_kind_to_string
0000000000023f30 T _as_agreement_new
0000000000023bf0 T _as_agreement_node_insert
```

samm-git: And the link to docs:

> The name of a symbol representing a function that conforms to standard C calling conventions is the name of the function with an underscore prefix.

https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/MachOTopics/1-Articles/executing_files.html

smcv: This should probably be called `appstream-glib.map.osx` or something, then - the presence or absence of the extra underscore is an OS ABI thing, not a compiler thing. (clang on Linux matches gcc on Linux, and doesn't prepend an underscore.)

(See also libtool macro `LT_SYS_SYMBOL_USCORE`.)

samm-git: @smcv file renamed.

hughsie: Thanks. If you rebase and squash down, then force push I think we're good. Thanks.

samm-git: @hughsie i squashed changes to 3 commits, let me know if it works.

PR #348, opened 2020-01-27
WIP: libasglib: Support all compression formats for AppStream XML repodata

Conan-Kudo:
Various Linux distributions are compressing AppStream XML repodata
with compression algorithms other than gzip. libappstream-glib needs
to recognize these so that they can be properly decompressed and used
accordingly.

### DIFF
--- a/libappstream-glib/as-format.c
+++ b/libappstream-glib/as-format.c
@@ -155,6 +155,12 @@ as_format_guess_kind (const gchar *filename)
 {
 	if (g_str_has_suffix (filename, ".xml.gz"))
 		return AS_FORMAT_KIND_APPSTREAM;
+	if (g_str_has_suffix (filename, ".xml.bz2"))
+		return AS_FORMAT_KIND_APPSTREAM;
+	if (g_str_has_suffix (filename, ".xml.xz"))
+		return AS_FORMAT_KIND_APPSTREAM;
+	if (g_str_has_suffix (filename, ".xml.zst"))
+		return AS_FORMAT_KIND_APPSTREAM;
 	if (g_str_has_suffix (filename, ".yml"))
 		return AS_FORMAT_KIND_APPSTREAM;
 	if (g_str_has_suffix (filename, ".yml.gz"))
--- a/libappstream-glib/as-utils.c
+++ b/libappstream-glib/as-utils.c
@@ -1275,6 +1275,24 @@ as_utils_install_filename (AsUtilsLocation location,
 			ret = as_utils_install_icon (location, filename, basename, destdir, error);
 			break;
 		}
+		tmp = g_strstr_len (basename, -1, "-icons.tar.bz2");
+		if (tmp != NULL) {
+			*tmp = '\0';
+			ret = as_utils_install_icon (location, filename, basename, destdir, error);
+			break;
+		}
+		tmp = g_strstr_len (basename, -1, "-icons.tar.xz");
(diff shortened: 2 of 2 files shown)


Conan-Kudo:
@hughsie This patch was written to resolve issues discovered in getting AppStream repodata to load in OpenMandriva. It will also affect ROSA and a few other distributions using non-gzip XML repodata.

(cc: @berolinux, @mikhailnov, @ximion)

Conan-Kudo:
@hughsie Ugh, this patch is not finished, just noticed the gzip hardcode in as-node.c...

berolinux:
Another identical use in as-yaml.c

Conan-Kudo:
@berolinux I'm not concerned about the YAML ones, that's only used by Debian and they're still using gzip.

ximion:
Well, that's actually not permitted by the specification... People can ship it over the wire as whatever they want, but in in app-info folders the files must either be raw data or GZip-compressed.
(on Debian, APT also changes the compression format to gzip explicitly in case it was something else)

Not sure if we should accommodate for distributions deliberately doing the wrong thing and implementing something that no compliant AppStream implementation supports.
Personally, I am very much against adding a wide array of different compression formats - if there was one more it should probably Zstd or LZ4 (XZ doesn't make much sense especially for appstream-glib, as that relies on fast decompression speed even more than libappstream does)

ximion:
Well, that's actually not permitted by the specification... ("The XML files might be compressed with GZip.") People can ship it over the wire as whatever they want, but in in app-info folders the files must either be raw data or GZip-compressed. The specification is very explicit on that matter (mentioning it twice, which is probably a style issue ^^).
On Debian, APT also changes the compression format to gzip explicitly in case it was something else before.

Not sure if we should accommodate for distributions deliberately doing the wrong thing and implementing something that no compliant AppStream implementation supports.
Personally, I am very much against adding a wide array of different compression formats - if there was one more it should probably be Zstd or LZ4 (XZ doesn't make much sense especially for appstream-glib, as that relies on fast decompression speed even more than libappstream does. And bz2 is pretty much dead for most purposes, as there are better options available now)

Conan-Kudo:
@ximion Then that means we should probably be decompressing this at the PackageKit backend level, before handing it to asglib?

Conan-Kudo:
Hmm, since this is the wrong kind of fix entirely, I'm closing this.

ximion:
> @ximion Then that means we should probably be decompressing this at the PackageKit backend level, before handing it to asglib?

Exactly. I am not against adding a new compression method if there's a benefit to it though. I wonder how many things would need to be adjusted to support that though, and whether that potential breakage is worth it... (ideally it would be only libappstream and libappstream-glib... @hughsie 's input is very welcome on this. Personally, I'm fine with gzip, it's fast to decompress and that's why we use it here in the first place)

Conan-Kudo:
@ximion So Fedora and Mageia currently ship repodata in gzip and zck, while OpenMandriva and ROSA ship in xz. I believe Arch is transitioning to zstd from gzip and xz.

ximion:
Debian ships it as gzip or xz, with xz being preferred (I just looked) and APT transparently recompresses AppStream data to be stored as gzip on-disk.

Conan-Kudo:
Right now, the only part of the DNF package manager stack that is aware of AppStream is PackageKit, so that's probably where the logic would need to be fixed...

PR #365, opened 2020-06-02
as-compose: Add checksum to icons file names

barthalion:
Flathub infrastructure aggressively caches screenshots and icons from appstream. It turned out to cause an issue for icons which do not include any hash in their names, meaning that when upstream project updated it, the old one was still displayed in the web store.

### DIFF
--- a/client/as-compose.c
+++ b/client/as-compose.c
@@ -11,6 +11,7 @@
 #include <appstream-glib.h>
 #include <glib.h>
 #include <glib/gi18n.h>
+#include <glib/gstdio.h>
 #include <stdlib.h>
 #include <locale.h>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


barthalion:
Converted to draft as currently flatpak does not resolve symlinks when extracting appstream.

hughsie:
Isn't the icon basename specified in the AppStream spec? I don't think we can just add the hash there.

barthalion:
Maybe I missed it, but I don't see it in spec. Wouldn't the symlink cover that case anyway?

barthalion:
I'm aware it's rather Flathub-specific change, I'll sort out some workaround from infrastructure side if it's not compliant with specification or too specific to our use case.

ximion:
The approach sounds wrong to me, like papering over a bug in Flathub's caching logic.
Caching icons is legit, they would just need to be updated if the AppStream metadata is also updated. And surely Flathub has to account for that in some way already, so it could generate its own hash or simply update the icons in lockstep then?

barthalion:
It's rather common way to handle cache invalidation without having to employ server-side logic (either by setting proper headers or purging data from CDN), and it's already there for screenshots via `mirror-screenshots`. In fact some websites never purge cached assets. It's not a hill I'm willing to die on though.

barthalion:
I assume nothing is going to happen about this so I'll just close it.

PR #366, opened 2020-06-09
as-content-rating: Bump the OARS/CSM mapping of sex-themes/intense

pwithnall:
This rationalises it with the OARS/CSM mapping of
sex-homosexuality/intense, and with many western societal norms. It is
against the laws and morals of various western countries to discriminate
on sexual orientation, so the mappings for sex-homosexuality/intense and
sex-themes/intense should be the same.

The mappings for the other values of sex-homosexuality and sex-themes
remain different, as their descriptions aren’t as comparable (for
example, for `mild`, the descriptions are “Indirect references to
homosexuality” vs “Provocative references or descriptions”).

The sex-homosexuality ratings category does need to continue to exist,
though, as it’s illegal in some countries to *not* discriminate on
sexual orientation. Those countries can provide and maintain their own
OARS/age mappings.

See https://gitlab.freedesktop.org/pwithnall/malcontent/-/issues/21.
This issue is now being solved in appstream-glib as the content rating
code has been moved there so it can be shared between malcontent and
gnome-software.

### DIFF
--- a/libappstream-glib/as-content-rating.c
+++ b/libappstream-glib/as-content-rating.c
@@ -1147,7 +1147,7 @@ static const struct {
 	{ "social-info",	OARS_1_0, 0, 0, 13, 13 },
 	{ "social-location",	OARS_1_0, 0, 13, 13, 13 },
 	/* v1.1 additions */
-	{ "sex-homosexuality",	OARS_1_1, 0, 10, 13, 18 },
+	{ "sex-homosexuality",	OARS_1_1, 0, 10, 13, 15 },
 	{ "sex-prostitution",	OARS_1_1, 0, 12, 14, 18 },
 	{ "sex-adultery",	OARS_1_1, 0, 8, 10, 18 },
 	{ "sex-appearance",	OARS_1_1, 0, 10, 10, 15 },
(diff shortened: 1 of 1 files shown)


pwithnall:
Once this is merged it will need to be backported to gnome-software and malcontent, since they still embed copies of the OARS/CSM mappings in lieu of bumping their hard dependency on appstream-glib.

libappstream-glib/as-content-rating.c
```diff
@@ -1134,7 +1134,7 @@ static const struct {
 	{ "drugs-narcotics",	OARS_1_0, 0, 12, 14, 17 },
 	{ "drugs-tobacco",	OARS_1_0, 0, 10, 13, 13 },
 	{ "sex-nudity",		OARS_1_0, 0, 12, 14, 14 },
-	{ "sex-themes",		OARS_1_0, 0, 13, 14, 15 },
+	{ "sex-themes",		OARS_1_0, 0, 13, 14, 18 },
```
hughsie, line 1137: The 15 there was deliberate after much debate with a child physiologist (aka my wife). We'd much rather lower the other one to match, rather than raise both.

pwithnall: Done!

pwithnall:
Copied to malcontent’s compatibility copy of this code here: https://gitlab.freedesktop.org/pwithnall/malcontent/-/merge_requests/71

PR #374, opened 2020-06-23
Don't ignore localized strings that are the same as original

igaldino:
Avoiding storing identical strings is a clever measure to keep the
appstream file small and clean, but it ignores translation differences
between languages and their specific locales.

Applications tend to "fallback" from missing locales by picking a
translation of the same language but from a different locale.
For instance, when missing a "pt_BR" translation, some apps will
pick the "pt" translation instead. That usually works but there
are some cases when it doesn't, such as for international words:
Brazilian Portuguese (pt_BR) tends to use them, while European
Portuguese (pt) has a translation for everything.

This way, "GNOME Boxes" gets translated to "Caixas GNOME" in
European Portuguese (pt) but the same "GNOME Boxes" name is
expected in Brazilian Portuguese (pt_BR).

This was initially reported as a Flatpak issue in
https://lists.freedesktop.org/archives/flatpak/2019-May/001578.html
Because OP was seeing the wrong translations in Flathub and
GNOME Software.

Original commit was done by @felipeborges in his personal branch at
github, commit 54f97aadcd2c00f3d95b472db6e4dda1c0c9865b

### DIFF
--- a/libappstream-glib/as-node.c
+++ b/libappstream-glib/as-node.c
@@ -1721,7 +1721,6 @@ as_node_get_localized (const AsNode *node, const gchar *key)
 	AsNodeData *data;
 	AsRefString *data_unlocalized;
 	AsRefString *xml_lang;
-	const gchar *data_localized;
 	GHashTable *hash = NULL;
 	AsNode *tmp;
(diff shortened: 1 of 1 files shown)


igaldino:
This is the resume of the work done in PR #307 by @felipeborges.

PR #376, opened 2020-07-18
validate: Make missing <releases> more obvious

jtojnar:
Previously, when <release> was not contained in <releases>, the following confusing error message was printed:

    • tag-missing           : <release> required

Now, the validation will print

    • tag-missing           : <releases> required

when <releases> is missing altogether and

    • tag-missing           : <release> required in <releases>

when the <releases> tag does not contain any <release>.


jtojnar:
This slightly improves that particular confusing warning but it is still far from ideal since it is too weak model at the moment. Ideally, we would use something like [XSD schema](https://github.com/ximion/appstream/issues/279) for doing the boring validity check and only write code for checking things that cannot be expressed in schemas.

PR #390, opened 2021-03-16
libappstream-builder: Fix paths for cached icons

pwithnall:
As per
https://www.freedesktop.org/software/appstream/docs/sect-AppStream-IconCache.html#spec-iconcache-location,
the size of an icon should be specified in its `width` and `height`
attributes, rather than as part of its name.

This was resulting in icons like this in the rpmfusion-free-33 appdata:
```
<icon type="cached" height="128" width="128">128x128/chromium-browser-privacy.png</icon>
```
which were then being resolved to the path:
```
/usr/share/app-info/icons/rpmfusion-free-33/128x128/128x128/chromium-browser-privacy.png
```
which contains a duplicate size subdirectory.

Signed-off-by: Philip Withnall <pwithnall@endlessos.org>

### DIFF
--- a/libappstream-builder/plugins/asb-plugin-font.c
+++ b/libappstream-builder/plugins/asb-plugin-font.c
@@ -662,17 +662,14 @@ asb_plugin_font_app (AsbPlugin *plugin, AsbApp *app,
 		}
 
 		/* add icon */
-		if (asb_context_get_flag (plugin->ctx, ASB_CONTEXT_FLAG_HIDPI_ICONS)) {
-			icon_filename = g_strdup_printf ("64x64/%s.png",
-							 as_app_get_id_filename (AS_APP (app)));
-		} else {
-			icon_filename = g_strdup_printf ("%s.png",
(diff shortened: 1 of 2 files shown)


pwithnall:
Note: The script rpmfusion uses to create its appdata files is [here](https://github.com/rpmfusion/rpmfusion-free-appstream-data/blob/master/update-appdata-rpmfusion-free.sh).

hughsie:
IIRC, we uses the filename prefix for older gnome-software versions that did not understand `width=` etc. It's probably something we don't care about for any supported gnome-software release now. It might be safer to merge this, AND to just get the basename in gnome-software rather than the full name so that new g-s can work with old and new metadata.

pwithnall:
> AND to just get the basename in gnome-software rather than the full name so that new g-s can work with old and new metadata.

The changes to gnome-software to be forwards- and backwards-compatible for this were merged yesterday: https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/653

PR #395, opened 2021-04-07
[WIP] Strip unknown tags

jonathonf:
An alternative approach to unknown tags within descriptions would be to strip them. This is a WIP PR to explore that approach.


hughsie:
Looks good, but it's screaming out for for self tests testing all the different permutations...

Issue #401, opened 2021-05-11
Validation should not accept multiple project_license elements


oturpe:
During [Fedora review of an application called qvge][review], a metainfo file that contains the following was created in attempt to describe the fact that qvge's binary contains parts for each of the listed licenses:

```
<project_license>MIT</project_license>
<project_license>BSD</project_license>
<project_license>LGPL-3.0</project_license>
```

My reading of the AppStream spec would be that there can be 0 or 1 of project_license elements, and an application that needs to respect all these licenses should be encoded as **MIT and BSD and LGPL-3.0**. But, Fedora's mandatory call to `appstream-util validate-relax` let the metainfo file in question pass. `validate-strict` does not complain about multiples either.

I think the validation should fail when there are multiple `project_license` elements. At least the reference implementation simply [ignores all but one][ref-impl], so having more is certainly an error.

[review]: https://bugzilla.redhat.com/show_bug.cgi?id=1913870
[ref-impl]: https://github.com/ximion/appstream/blob/55973af9fcf145cbf85b74f24ff61ee6ea922c9d/src/as-component.c#L3749

This is the full file that should fail validation, but does not:
```
<?xml version="1.0" encoding="UTF-8"?>
<!-- Copyright 2016-2020 Ars L. Masiuk <ars.masiuk@gmail.com> -->
<component type="desktop">
  <id>qvge.desktop</id>
  <metadata_license>MIT</metadata_license>
  <project_license>MIT</project_license>
  <project_license>BSD</project_license>
  <project_license>LGPL-3.0</project_license>
  <name>Qt visual graph editor</name>
  <summary>View and manipulate small till middle-sized graphs.</summary>
  <description>
    <p>
      Application for viewing and manipulating small till middle-sized graphs. It supports several formats (GraphML, GEXF, GML, GraphViz/Dot).
    </p>
  </description>
  <url type="homepage">https://github.com/ArsMasiuk/qvge</url>
  <url type="donation">https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&amp;hosted_button_id=Z35EHHJ3729GG&amp;source=url</url>
  <screenshots>
    <screenshot type="default">
      <image>https://user-images.githubusercontent.com/19762856/85934275-2a985480-b8e1-11ea-81a9-a5b81f3365e2.PNG</image>
      <caption>Screenshot of the main window</caption>
    </screenshot>
  </screenshots>
[...]

PR #402, opened 2021-05-13
Add validation for duplicate xml elements

oturpe:
Draft fix of #401 

As an example, duplicate checking for two elements is added. Using this simple system, checking all elements requires replicating the code for each checked field. Solution with less replication would require a new data structure and passing it around: instead of enumerated problems like `AS_APP_PROBLEM_DUPLICATE_PROJECT_LICENSE`, some kind of generic duplicate problem entry that takes element name as input.

Posting this draft for review, because I am really new to this codebase and cannot say what would be best way to proceed:
1. Replicate the simple solution for all fields that can only appear once
2. Develop the generic solution
3. Only check `project_license` duplicates, as that is what was reported in #401 and there the root of the problem is that even if a project may be under multiple licenses, there can only be one license field and separate syntax for encoding everything in there must be used.

Comments welcome.


libappstream-glib/as-app-validate.c
```diff
@@ -1525,6 +1525,16 @@ as_app_validate (AsApp *app, guint32 flags, GError **error)
 					     AS_PROBLEM_KIND_TAG_INVALID,
 					     "<updatecontact> should be <update_contact>");
 		}
+		if ((problems & AS_APP_PROBLEM_DUPLICATE_PROJECT_LICENSE) > 0) {
+            ai_app_validate_add(helper,
+            	         AS_PROBLEM_KIND_TAG_INVALID,
+            	         "<project_license> was duplicated");
```
hughsie, line 1531: intenting looks funky, 8 width tabs please

libappstream-glib/as-app.c
```diff
@@ -5154,15 +5154,21 @@ as_app_node_parse_child (AsApp *app, GNode *n, guint32 flags,
 			priv->problems |= AS_APP_PROBLEM_TRANSLATED_LICENSE;
 			break;
 		}
+		if (priv->project_license != NULL) {
+     		priv->problems |= AS_APP_PROBLEM_DUPLICATE_PROJECT_LICENSE;
+	    }
```
hughsie, line 5159: indenting too

hughsie: also no braces required

libappstream-glib/as-app.c
```diff
@@ -5154,15 +5154,21 @@ as_app_node_parse_child (AsApp *app, GNode *n, guint32 flags,
...
-	/* <project_license> */
+	/* <metadata_license> */
 	case AS_TAG_METADATA_LICENSE:
 		if (as_node_get_attribute (n, "xml:lang") != NULL) {
 			priv->problems |= AS_APP_PROBLEM_TRANSLATED_LICENSE;
 			break;
 		}
+		if (priv->metadata_license != NULL) {
+     		priv->problems |= AS_APP_PROBLEM_DUPLICATE_METADATA_LICENSE;
+	    }
```
hughsie, line 5171: same

hughsie:
A few nits, but looks basically what I would have done, so good job.

oturpe:
Formatting fixed. A test case failed, apparently the "bad file" test already had a test file that had these problems, now the asserts need to be fixed. I will see to that soon.

oturpe:
Tests added. In absence of further comments about which elements should be tested for duplicates, I left the scope as it was initially: only the license fields.

oturpe:
Travis CI build failed, but to me that looks like an unrelated issue, as if something is generated without strict ordering, but is checked against fixed expectation:

```
Bail out! As:ERROR:../libappstream-builder/asb-self-test.c:524:asb_test_context_func: assertion failed (error == NULL):
--- /tmp/b	2021-05-17 19:42:30.984710347 +0000
+++ /tmp/a	2021-05-17 19:42:30.984710347 +0000
@@ -24,8 +24,8 @@
 <name>App</name>
 <summary>A test application</summary>
 <description><p>Long description goes here.</p></description>
-<icon type="cached" height="128" width="128">app.png</icon>
 <icon type="cached" height="64" width="64">app.png</icon>
+<icon type="cached" height="128" width="128">app.png</icon>
 <categories>
 <category>Profiling</category>
 <category>System</category>
 (-g-type-private--GTypeFlags, 0)
```

hughsie:
> but to me that looks like an unrelated issue

Agreed! Could I be cheeky and ask you to add the fix for that to this branch please? It should just be reordering the expected lines in asb-self-test.c. Thanks!

oturpe:
That error is now fixed, but another one was hidden behind it:

```
Bail out! As:ERROR:../libappstream-builder/asb-self-test.c:628:asb_test_context_func: assertion failed: (g_file_test ("/tmp/asbuilder/temp/icons/64x64/app.png", G_FILE_TEST_EXISTS))

--- stderr ---
**
As:ERROR:../libappstream-builder/asb-self-test.c:628:asb_test_context_func: assertion failed: (g_file_test ("/tmp/asbuilder/temp/icons/64x64/app.png", G_FILE_TEST_EXISTS))
-------
```

hughsie:
> That error is now fixed, but another one was hidden behind it:

Okay, I can look at that later. Thanks for the fixes!

PR #403, opened 2021-06-11
Support em/code tags

philmmanjaro:
Some appstream-data packages add \<em\> and \</em\> or \<code\> and \</code\> to the files.
Not all package manager can handle that. An example would be Pamac from Manjaro

### DIFF
--- a/libappstream-glib/as-node.c
+++ b/libappstream-glib/as-node.c
@@ -555,6 +555,8 @@ typedef struct {
 	AsNode			*current;
 	AsNodeFromXmlFlags	 flags;
 	const gchar * const	*locales;
+	guint8			 is_em_text;
+	guint8			 is_code_text;
 } AsNodeToXmlHelper;
 
 /**
@@ -603,6 +605,16 @@ as_node_start_element_cb (GMarkupParseContext *context,
 	AsNodeData *data_parent;
 	AsNode *current;
 	guint i;
+
+	/* do not create a child node for em and code tags */
+	if (g_strcmp0 (element_name, "em") == 0) {
+		helper->is_em_text = 1;
+		return;
+	}
+	if (g_strcmp0 (element_name, "code") == 0) {
+		helper->is_code_text = 1;
+		return;
+	}
 
 	/* check if we should ignore the locale */
 	data = g_slice_new0 (AsNodeData);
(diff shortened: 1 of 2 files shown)


libappstream-glib/as-node.c
```diff
@@ -604,6 +606,16 @@ as_node_start_element_cb (GMarkupParseContext *context,
...
+	/* do not create a child node for em and code tags */
+	if (g_strcmp0 (element_name, "em") == 0) {
+		helper->is_em_text = 1;
+		return;
+	} else if (g_strcmp0 (element_name, "code") == 0) {
+		helper->is_code_text = 1;
+		return;
+	}
+
+
```
hughsie, line 618: just one newline pls

libappstream-glib/as-node.c
```diff
@@ -555,6 +555,8 @@ typedef struct {
 	AsNode			*current;
 	AsNodeFromXmlFlags	 flags;
 	const gchar * const	*locales;
+	guint8 is_em_text;
+	guint8 is_code_text;
```
hughsie, line 559: need to align with the other vars pls

hughsie:
I think we need something added to the self tests as well, no?

libappstream-glib/as-node.c
philmmanjaro: removed the one double newline

libappstream-glib/as-node.c
philmmanjaro: updated

philmmanjaro:
Tests are now added.

libappstream-glib/as-self-test.c
```diff
@@ -2928,6 +2933,16 @@ as_test_node_xml_func (void)
...
+	/* support em and code tags */
+	root = as_node_from_xml (valid_em_code, 0, &error);
+	g_assert_no_error (error);
+	g_assert (root != NULL);
+
+	n2 = as_node_find (root, "description/p");
+	g_assert (n2 != NULL);
+	printf ("<%s>\n", as_node_get_data (n2));
+	g_assert_cmpstr (as_node_get_data (n2), ==, "It now also supports<em>em</em> and <code>code</code> tags.");
+
```
hughsie, line 2945: do you need a as_node_unref (root); here?

hughsie:
One little nit, and then please squash all the commits into one commit and force push to this branch. Thanks.

PR #407, opened 2021-07-23
Update current newest version in AsStore

### LINKED ISSUE #406: Increase AS_API_VERSION_NEWEST
According to https://www.freedesktop.org/software/appstream/docs/chap-CollectionData.html#spec-asxml-general the current version is 0.9

tintou:
We are now in version 0.14

Closes #406

Signed-off-by: Corentin Noël <corentin.noel@collabora.com>

### DIFF
--- a/libappstream-glib/as-store.c
+++ b/libappstream-glib/as-store.c
@@ -35,7 +35,7 @@
 #include "as-yaml.h"
 #include "as-store-cab.h"
 
-#define AS_API_VERSION_NEWEST	0.8
+#define AS_API_VERSION_NEWEST	0.14
 
 typedef enum {
 	AS_STORE_PROBLEM_NONE			= 0,


ximion:
I wonder: Are there any integer comparisons made with this variable? Because `0.14 < 0.9` - for this reason, libappstream uses an enum for this, so you can still easily check whether the current target API version is > a tested version.
If there's no direct comparisons made with this definition, this change is fine.

tintou:
Hmm, indeed there is indirectly here: https://github.com/hughsie/appstream-glib/blob/6d6b8f488e4cd072818af037978e44fe61bf3415/libappstream-glib/as-app.c#L4834 that compares it.

hughsie:
@tintou could you do a PR to change that comparison to use as_utils_vercmp() please. Thanks!

ximion:
@hughsie `AS_API_VERSION_NEWEST` would need to become a string constant then - is this API exported anywhere? Libappstream "cheats" an uses an enum for this ^^ ( https://github.com/ximion/appstream/blob/master/src/as-context.h#L49 )

PR #409, opened 2021-07-25
meson: Check --version-script linker option before using.

Biswa96:
mingw clang does not support --version-script linker option.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -2,8 +2,10 @@ project('appstream-glib', 'c',
   version : '0.7.19',
   license : 'LGPL-2.1+',
   default_options : ['warning_level=1', 'c_std=c99'],
-  meson_version : '>=0.37.0'
+  meson_version : '>=0.46.0'
 )
+
+cc = meson.get_compiler('c')
 
 as_version = meson.project_version()
 varr = as_version.split('.')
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -2,9 +2,11 @@ project('appstream-glib', 'c',
   version : '0.7.19',
   license : 'LGPL-2.1+',
   default_options : ['warning_level=1', 'c_std=c99'],
-  meson_version : '>=0.37.0'
+  meson_version : '>=0.46.0'
```
hughsie, line 5: do we really need this new version?

Biswa96: With the older meson version requirement, this warnings are show:

```
WARNING: Project specifies a minimum meson_version '>=0.37.0' but uses features which were added in newer versions:
 * 0.46.0: {'compiler.get_supported_link_arguments_method', 'compiler.has_multi_link_argument', 'compiler.has_link_argument'}
 * 0.50.0: {'install arg in configure_file'}
```

meson 0.46.0 was released in Apr 23, 2018

hughsie: > meson 0.46.0 was released in Apr 23, 2018

Fair point.

PR #410, opened 2021-08-02
Consider AppStream version as a string

tintou:
Always store the AppStream metadata version as a string and compare it
like any other version.
This allows to have 0.10 > 0.8 for instance.


hughsie:
Don't you need to add the new symbols to appstream-glib.map?

tintou:
It looks like adding it to appstream-glib.map isn't necessary:
```
{
global:
    as_*;
local:
    *;
};
```

hughsie:
> as_*;

dd0

PR #415, opened 2021-08-10
Don't emit legacy component type values

ximion:
Emitting the old names was likely kept to not create any breakage and provide full backwards compatibility between different versions. However, asglib has supported the new names for years now and emitting the older ones actually causes problems now, in the case of `webapp`: https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/899
This change also makes asglib emit `desktop-application` instead of `desktop`, which likely doesn't fix any real-world issues, as we'll keep the `desktop` alias forever, but it's nice to emit the full name for consistency.

### DIFF
--- a/libappstream-glib/as-app.c
+++ b/libappstream-glib/as-app.c
@@ -138,7 +138,7 @@ const gchar *
 as_app_kind_to_string (AsAppKind kind)
 {
 	if (kind == AS_APP_KIND_DESKTOP)
-		return "desktop";
+		return "desktop-application";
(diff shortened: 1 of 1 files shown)


hughsie:
I reverted this, it broke the self tests in lots of places.

ximion:
I would argue that these tests should probably be fixed ;-)
At least "webapp" was never standardized and libappstream will not recognize it ("desktop" is a different case, that one will probably be supported until the universe dies, way beyond AppStream 1.0 at least).

PR #432, opened 2022-03-09
Fix extracting relative symlinks

Vogtinator:
https://github.com/hughsie/appstream-glib/commit/ebdefa2745292af75ac39b1b69f7b25aaeb35838 tried to fix handling of relative symlinks in archives, but that
only addressed targets starting with ".."/. Fix it properly by using
asb_utils_sanitise_path only for absolute paths, keep relative paths as-is.

Simplify code for handling absolute symlinks by just prepending the root
instead of converting it to a relative link.

Hardlink targets are returned as paths relative to the archive root, so the
sanitization code is correct there.


hughsie:
Back a bit; what bug does this solve?

Vogtinator:
https://bugzilla.opensuse.org/show_bug.cgi?id=1196459 is the downstream commit. The affected package has 

```
lrwxrwxrwx 1 root root 29 19. Sep 05:14 /usr/share/icons/hicolor/32x32/apps/simplescreenrecorder.png -> simplescreenrecorder-idle.png
```

libappstream-builder extracted that as

```
lrwxrwxrwx 1 fabian users 47  9. Mär 15:57 tmp/simplescreenrecorder-0.4.3-1.11/usr/share/icons/hicolor/32x32/apps/simplescreenrecorder.png -> ../../../../../../simplescreenrecorder-idle.png
```

i.e. `simplescreenrecorder-idle.png` "sanitized" to `/simplescreenrecorder-idle.png` and converted to the relative `../../../../../../simplescreenrecorder-idle.png`.

The missing icon file triggers a veto and ultimately the package is missing from the generated collection.

PR #433, opened 2022-03-09
Create icons/64x64 folder also without --enable-hidpi

Vogtinator:
The icons are written into the folder, so make sure it exists. Without this,
appstream-builder fails if the temporary directory didn't have icons/64x64 yet.

I guess that previously icons didn't have width/height set, so they were written into the `icons` folder directly instead of the `64x64` subfolder.

### DIFF
--- a/libappstream-builder/asb-context.c
+++ b/libappstream-builder/asb-context.c
@@ -510,6 +510,7 @@ asb_context_setup (AsbContext *ctx, GError **error)
 	g_autofree gchar *icons_dir = NULL;
 	g_autofree gchar *screenshot_dir1 = NULL;
 	g_autofree gchar *screenshot_dir2 = NULL;
+	g_autofree gchar *icons_dir_lodpi = NULL;
 
 	/* required stuff set */
 	if (priv->origin == NULL) {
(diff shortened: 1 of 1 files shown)


libappstream-builder/asb-context.c
```diff
@@ -565,12 +565,14 @@ asb_context_setup (AsbContext *ctx, GError **error)
 	/* icons is nuked; we can re-decompress from the -icons.tar.gz */
 	if (!asb_utils_ensure_exists (priv->icons_dir, error))
 		return FALSE;
+
+	g_autofree gchar *icons_dir_lodpi = NULL;
```
hughsie, line 569: this needs to be at the start of the block like the others

libappstream-builder/asb-context.c
```diff
@@ -565,12 +565,14 @@ asb_context_setup (AsbContext *ctx, GError **error)
...
+	icons_dir_lodpi = g_build_filename (priv->icons_dir, "64x64", NULL);
+	if (!asb_utils_ensure_exists (icons_dir_lodpi, error))
+		return FALSE;
+
 	if (priv->flags & ASB_CONTEXT_FLAG_HIDPI_ICONS) {
 		g_autofree gchar *icons_dir_hidpi = NULL;
-		g_autofree gchar *icons_dir_lodpi = NULL;
-		icons_dir_lodpi = g_build_filename (priv->icons_dir, "64x64", NULL);
-		if (!asb_utils_ensure_exists (icons_dir_lodpi, error))
-			return FALSE;
```
hughsie, line 573: doesn't this change the logic? i.e. we'd put the icons in priv->icons_dir -- unless we used HIDPI and then we'd put them in priv->icons_dir/64x64 and priv->icons_dir/128x128

libappstream-builder/asb-context.c
Vogtinator: Indeed

libappstream-builder/asb-context.c
hughsie: I'm wondering if we should just remove enable-hidpi completely and just always write 64x64 and 128x128 icons. It would certainly make the code a lot simpler...

Vogtinator: > I'm wondering if we should just remove enable-hidpi completely and just always write 64x64 and 128x128 icons. It would certainly make the code a lot simpler...

Aggreed.

ximion: Shouldn't the 64x64px icons always be available anyway? That size is pretty much the only size required to be present by the specification, with all other icon dimensions being optional.

hughsie: Dude, this is for super old clients that expected the icons in /usr/share/app-info/icons/$origin/ i.e. without the 64x64 suffix. I think we can drop support for that now.

hughsie:
Alternate proposal in https://github.com/hughsie/appstream-glib/pull/434

PR #438, opened 2022-03-30
Add vcs-browser and contribute URL type

JakobDev:
This 2 new URL types [were added to AppStream](https://github.com/ximion/appstream/pull/392). This PR adds Support for them in appstream-glib.


JakobDev:
@hughsie Thanks. Could you please create a new Release, so it's possible to use this new types on Flathub?

hughsie:
I wasn't planning a release any time soon, but I can put it on the TODO for next week.

PR #441, opened 2022-06-07
Port from libsoup to libcurl

hughsie:
The former bumped ABI, and all sorts of crazy happens when you link in
libappstream-glib into a process with the 'other' ABI.

It seems the universe has settled on curl as a dep; do the same.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ copy. To do the latter just do:
 
     dnf install docbook-utils gettext-devel glib-devel \
                 gobject-introspection-devel gperf gtk-doc gtk3-devel \
-                json-glib-devel libarchive-devel libsoup-devel \
+                json-glib-devel libarchive-devel libcurl-devel \
                 libstemmer-devel libuuid-devel libyaml-devel \
                 meson rpm-devel
     mkdir build && cd build
--- a/client/as-util.c
+++ b/client/as-util.c
@@ -18,7 +18,7 @@
 #include <appstream-glib.h>
 #include <archive_entry.h>
 #include <archive.h>
-#include <libsoup/soup.h>
+#include <curl/curl.h>
 #include <locale.h>
 #include <stdlib.h>
 
@@ -38,6 +38,7 @@ typedef struct {
 	GMainLoop		*loop;
 	GCancellable		*cancellable;
 	AsProfile		*profile;
+	CURL			*curl;
 } AsUtilPrivate;
 
 typedef gboolean (*AsUtilPrivateCb)	(AsUtilPrivate	*util,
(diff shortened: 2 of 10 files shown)


nanonyme:
Sounds fine to me. It has been repeatedly raised libcurl doesn't support autoproxy but I'm not sure if it matters in modern world. Evaluating WPAD scripts is a significant attack vector.

ximion:
I think this is a good choice - I switched from Soup to Curl in libappstream a long time ago as well, and that was a great choice in retrospect.

hughsie:
> This will be a big issue for our company

Ohh? What are you using appstream-glib for out of interest?

smcv:
> Please support libsoup as libcurl does not support pac proxy. This will be a big issue for our company

If this is a problem for you, I think the way to use a PAC-detected proxy with libcurl would be to open a pull request that does something like this for each outgoing request:

* use `g_proxy_resolver_lookup()` to get the proxy/proxies to use from GLib
* go through the list of proxies from GLib, find the first one that is understood and translate it into libcurl syntax:
    * `http://` can be used directly (I think)
    * `socks:` -> either `socks4:// `, `socks4a:// `, `socks5:// ` or `socks5h:// ` whichever is most appropriate
    * `direct://` -> empty string
* `curl_easy_setopt(curl, CURLOPT_PROXY, "http://proxy:80");` replacing `"http://proxy:80"` with the proxy that was translated from GLib to libcurl syntax

mcatanzaro:
BTW I see you use GProxyResolver and curl with fwupd already, so either (a) you've already figured this out and can copy/paste, or (b) you're doing it wrong there.

mcatanzaro:
> BTW I see you use GProxyResolver and curl with fwupd already, so either (a) you've already figured this out and can copy/paste, or (b) you're doing it wrong there.

And looks like it's _mostly_ (a). Very quick check of [this code](https://github.com/fwupd/fwupd/blob/09c703638f16b032c0475cb270bc272b2d46a216/libfwupd/fwupd-client.c#L626) looks like mostly correct usage to me. I don't think you're handling SOCKS properly, though. You might need to translate the SOCKS URLs as suggested my Simon above.

mcatanzaro:
> socks: -> either socks4:// , socks4a:// , socks5:// or socks5h:// whichever is most appropriate

Looks like GProxyResolver tries to do this already. If we're using the libproxy-based GProxyResolver, it translates socks:// from libproxy into socks5://, then socks4a://, then socks4://. It doesn't know about socks5h:// though. But the GProxyResolver that uses GNOME settings, and the one that uses environment variables, just pass along whatever is configured. I suppose being prepared for the unexpected certainly wouldn't hurt.

smcv:
> If we're using the libproxy-based GProxyResolver, it translates socks:// from libproxy into socks5://, then socks4a://, then socks4://

That sounds like the documentation for `g_proxy_resolver_lookup()` either disagrees with the implementation or is misleading, because it talks about how "\<protocol> could be http, rtsp, socks or other proxying protocol".

mcatanzaro:
> That sounds like the documentation for g_proxy_resolver_lookup() either disagrees with the implementation or is misleading, because it talks about how "<protocol> could be http, rtsp, socks or other proxying protocol".

Drat, yes. :/ It's arguably OK if we interpret it as "expect anything, yolo," but misleading at the very least.

mcatanzaro:
So the real documentation is in `GSimpleProxyResolver` which specifies the behavior explicitly: "if proxy starts with "socks://", #GSimpleProxyResolver will treat it as referring to all three of the socks5, socks4a, and socks4 proxy types."

The more generic `GProxyResolver` can theoretically do anything, I suppose. But in practice, reasonable implementations will match GSimpleProxyResolver.

hughsie:
Fixed in https://github.com/hughsie/appstream-glib/pull/443 ?

PR #443, opened 2022-06-16
Use GProxyResolver to auto-detect the correct proxy

hughsie:


hughsie:
@mcatanzaro

mcatanzaro:
CC @janbrummer

Issue #445, opened 2022-07-15
as-self-test fails since 0.8.0


heftig:
The `as-self-test` fails since 0.8.0:

```
<It now also supports<em>em</em><code> and </code><code>code</code> tags.>
Bail out! As:ERROR:../appstream-glib/libappstream-glib/as-self-test.c:2944:as_test_node_xml_func: assertion failed (as_node_get_data (n2) == "It now also supports<em>em</em> and <code>code</code> tags."): ("It now also supports<em>em</em><code> and </code><code>code</code> tags." == "It now also supports<em>em</em> and <code>code</code> tags.")
stderr:
**
As:ERROR:../appstream-glib/libappstream-glib/as-self-test.c:2944:as_test_node_xml_func: assertion failed (as_node_get_data (n2) == "It now also supports<em>em</em> and <code>code</code> tags."): ("It now also supports<em>em</em><code> and </code><code>code</code> tags." == "It now also supports<em>em</em> and <code>code</code> tags.")
```

Bisection identified f939f14774618fd07d7019e9d0c86e1e1ae5642a as the first broken commit.

`meson setup` + `compile` log: [appstream-glib-0.8.0-1-x86_64-build.log](https://github.com/hughsie/appstream-glib/files/9121597/appstream-glib-0.8.0-1-x86_64-build.log)
`meson test` log: [appstream-glib-0.8.0-1-x86_64-check.log](https://github.com/hughsie/appstream-glib/files/9121588/appstream-glib-0.8.0-1-x86_64-check.log)

hughsie:
It worked here (and continues to work); can you debug a bit please?

heftig:
The new fields in the `AsNodeToXmlHelper` struct aren't being initialized.

(Also, they should probably be bitfields for consistency.)

hughsie:
Can you do a pr please?

PR #446, opened 2022-07-15
Improve handling of <em> and <code> tags

heftig:
Three commits:
- [Properly initialize is_{em,code}_text fields](https://github.com/hughsie/appstream-glib/commit/9c0b92916dbadc16e66a4b782cfc773a8a041d35)
  Fixes: https://github.com/hughsie/appstream-glib/issues/445
- [trivial: Turn is_{em,code}_text fields into bitfields](https://github.com/hughsie/appstream-glib/commit/cbfa4a528c3e6eaee22475a98ee092269cbb0b95)
- [Improve handling of \<em\> and \<code\> tags](https://github.com/hughsie/appstream-glib/commit/bdebc8abfe0dbf5d9d5488af365b291f4b103ed0)

  This is still not great code but at least somewhat an improvement. Tests were expanded to showcase the new behavior.

  I think, ideally, we would append opening/closing tags to the ancestor `p` or `li` node's cdata as soon as we encounter the start/end of an `em` or `code` element. This would then also handle empty elements correctly.

### DIFF
--- a/libappstream-glib/as-node.c
+++ b/libappstream-glib/as-node.c
@@ -555,8 +555,8 @@ typedef struct {
 	AsNode			*current;
 	AsNodeFromXmlFlags	 flags;
 	const gchar * const	*locales;
-	guint8			 is_em_text;
-	guint8			 is_code_text;
+	guint8			 is_em_text:1;
+	guint8			 is_code_text:1;
 } AsNodeToXmlHelper;
 
 /**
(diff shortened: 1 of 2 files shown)


libappstream-glib/as-node.c
```diff
@@ -844,6 +864,7 @@ as_node_from_xml_internal (const gchar *data, gssize data_sz,
 	helper.flags = flags;
 	helper.current = root;
 	helper.locales = g_get_language_names ();
+	helper.is_em_text = helper.is_code_text = 0;
```
hughsie, line 867: I think it would be better to prevent future incidents to just have `AsNodeToXmlHelper helper = {0}` everywhere it's allocated.

heftig: Changed as requested.

hughsie (approved):

hughsie:
Thanks!

PR #450, opened 2022-07-22
Initialize proxy resolver in appstream-validate

### LINKED ISSUE #449: as-app-validate.c fails to initialize GProxyResolver before using it
as-app-validate.c fails to initialize GProxyResolver before using it, resulting in criticals:

https://gitlab.gnome.org/GNOME/glib/-/issues/2697

mcatanzaro:
Oops, trying to use it would just result in criticals.

Fixes #449

NOTE: this is untested: I didn't even attempt to build it. I've provided this because it was requested in #449. Will probably work.


hughsie:
Great, thanks.

PR #451, opened 2022-07-28
po: Add Georgian translation

NorwayFun:
po: Add Georgian translation


hughsie:
I think you need to add this to LINGUAS too, no?

PR #452, opened 2022-07-28
Update LINGUAS: add Georgian

NorwayFun:


hughsie:
Can you combine the PRs into one please.

PR #460, opened 2022-10-08
Tell CURL to follow http redirect

hfiguiere:
- This is issue https://github.com/flathub/org.flatpak.Builder/issues/108 and https://github.com/flathub/org.flatpak.Builder/issues/112

Explanation: the switch to curl seems to have missed the fact that by default curl doesn't follow redirects

And flatpak-builder with `--mirror-screenshots-url` did fail when screenshot were hosted (but no only) on github.


hughsie (approved):

hfiguiere:
Any chance to get 0.8.2 out with this fix?

hfiguiere:
There is a second place with a similar bug.

Issue #419, opened 2021-10-14
Don’t add default icons to input sources when composing appstream


pwithnall:
It’s causing input sources to appear with the wrong icon in gnome-software. See https://gitlab.gnome.org/GNOME/gnome-software/-/issues/1483

There’s no need to add a default icon to input sources because [it’s not required by the spec](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-InputMethod.html#spec-imdata-filespec). If client applications like gnome-software want to always show an icon, let them choose their own defaults.

This may also apply to codecs, fonts and other non-application components; I haven’t checked.

pwithnall:
(I have no intention of following this up further: just filing the bug so some interested party from Fedora can follow it up, or at least all the conversation about it can happen in one place.)

pwithnall:
I wrote the trivial patch: https://github.com/hughsie/appstream-glib/pull/462

PR #462, opened 2022-10-10
asb-plugin-appdata: Never add default icons to firmware, IMs or codecs

pwithnall:
Consumers of the appdata file are perfectly capable of adding their own default icons, and the specification doesn’t require icons to be present for any of these component types.

If appstream-builder provides a default icon, software centres cannot work out whether it’s provided by the component authors or by appstream-builder. This means software centres cannot override it with their own desktop-specific default icons.

Signed-off-by: Philip Withnall <pwithnall@endlessos.org>

See: https://gitlab.gnome.org/GNOME/gnome-software/-/issues/1483
Fixes: #419

### DIFF
--- a/libappstream-builder/plugins/asb-plugin-appdata.c
+++ b/libappstream-builder/plugins/asb-plugin-appdata.c
@@ -138,33 +138,14 @@ asb_plugin_process_filename (AsbPlugin *plugin,
 	/* fix up various components as required */
 	if (asb_context_get_flag (plugin->ctx, ASB_CONTEXT_FLAG_ADD_DEFAULT_ICONS)) {
 
-		/* add icon for firmware */
-		if (as_app_get_kind (AS_APP (app)) == AS_APP_KIND_FIRMWARE) {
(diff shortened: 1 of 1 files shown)


pwithnall:
Untested

hughsie:
I think we did this explicitly for old versions of gnome-software -- which we probably don't care about now. @ximion do you need this fallback for any of the KDE software centers, or do you do anything similar in appstream? Thanks.

ximion:
No, appstream never did this, and KDE software centers never relied on it - KDE Discover does show a pretty bad generic "package" icon though, it would be neat if it at least showed different ones for console-apps, firmware and drivers. That's a KDE problem though :-)

If a GNOME Software version relied on this, it must have been so old that it was never an issue for Debian at least (certainly not for any of its current stable releases, with GNOME Software 3.30.x being the oldes (GS 3.22.x also exists in oldoldstable, but people should really have migrated away from that by now)), so this feels pretty safe to drop :-)

pwithnall:
Thanks both :)

PR #464, opened 2022-12-29
Include screenshot size in all error messages

suve:
Screenshot validation includes checking the image size, i.e:
- whether the width isn't too small
- whether the height isn't too small
- whether the width isn't too large
- whether the height isn't too large

Currently, when the first check fails, the error message includes the actual screenshot width. This is not the case for other checks - they do not contain any information about the dimensions. This patch fixes this inconsistency by making all four checks include the relevant information in their error message.

### DIFF
--- a/libappstream-glib/as-app-validate.c
+++ b/libappstream-glib/as-app-validate.c
@@ -555,8 +555,8 @@ ai_app_validate_image_check (AsImage *im, AsAppValidateHelper *helper)
 	if (screenshot_height < ss_size_height_min) {
 		ai_app_validate_add (helper,
 				     AS_PROBLEM_KIND_ATTRIBUTE_INVALID,
-				     "<screenshot> height too small [%s] minimum is %upx",
-				     url, ss_size_height_min);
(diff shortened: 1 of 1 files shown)


hughsie:
Medium term I think we should point people at appstreamcli. Does this PR still pass ninja test?

suve:
> Does this PR still pass ninja test?
```
[65/66] Running all tests.
1/2 asb-self-test        OK              0.09s
2/2 as-self-test         OK              0.19s

Ok:                 2   
Expected Fail:      0   
Fail:               0   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0   
```

> Medium term I think we should point people at `appstreamcli`.

[Fedora's packaging guidelines on AppStream files](https://docs.fedoraproject.org/en-US/packaging-guidelines/AppData/) require running `appstream-util validate-relax` on said files. I guess long term we'd want to switch to `appstreamcli` here, as well?

hughsie:
> I guess long term we'd want to switch to appstreamcli here, as well?

I think so. Thanks!

PR #468, opened 2023-06-18
Add "get-latest-version" util command

hadess:
This gets the latest version specified in an appdata file, which, combined with utilities such as check-news.sh[1], can be used to check whether the date specified in the appdata's release information is correct.

See https://gitlab.gnome.org/GNOME/evince/-/issues/1950

[1]: https://gitlab.gnome.org/GNOME/totem/-/blob/master/check-news.sh

### DIFF
--- a/client/as-util.c
+++ b/client/as-util.c
@@ -2861,6 +2861,88 @@ as_util_validate_files (gchar **filenames,
 }
 
 static gboolean
+as_util_get_latest_version_file (gchar   *filename,
+				 GError **error)
+{
+	g_autoptr(AsApp) app = NULL;
+	AsRelease *release = NULL;
+	AsFormatKind kind;
+
+	/* is AppStream */
+	kind = as_format_guess_kind (filename);
+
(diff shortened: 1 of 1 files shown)


hughsie:
This makes total sense, but I think it should be added to the https://github.com/ximion/appstream project instead; appstream-glib is in heavy maintenance mode now and we're not really taking on new features. I think most things have been ported, or am porting to the newer libappstream library maintained by @ximion

hadess:
> This makes total sense, but I think it should be added to the https://github.com/ximion/appstream project instead; appstream-glib is in heavy maintenance mode now and we're not really taking on new features. I think most things have been ported, or am porting to the newer libappstream library maintained by @ximion

Would have been nice if the README mentioned it :/

hadess:
Filed https://github.com/ximion/appstream/issues/502 against appstream until I get the time to reimplement this once more.

hadess:
> Would have been nice if the README mentioned it :/

Done in https://github.com/hughsie/appstream-glib/pull/469

PR #470, opened 2023-07-22
as-app: Print name of unknown app metadata tags

sidt4:
Currently, we just get a "XML data contains unknown tag" error message, without the actual tag name when running `"appstream-util validate-strict"`.

```
$ appstream-util validate-strict --nonet org.gnome.Nautilus.appdata.xml org.gnome.Nautilus.appdata.xml: FAILED:
..
• tag-invalid           : XML data contains unknown tag
..
Validation of files failed
```

With this commit, we now display a warning for each unknown tag.

```
(appstream-util:289727): As-WARNING **: 14:12:10.446: Unknown tag: <recommends>
```


sidt4:
Okay.

PR #498, opened 2025-02-27
asb-plugin-font: Add tentative symbol text for N'Ko, Yi script, Amharic, and Inuktitut

tagoh:
It may be too long for text icon but better than nothing so far. I'm not a native speaker of them. we could replace it later.


tagoh:
I'm not adding sample text strings for them. I have no idea what is the better sentences.

tagoh:
I added something for Amharic and Inuktitut too. This is something we can do at the moment for missing icons/screenshots issues for Noto fonts.  I can still see warnings but it is hard to take care of them because those scripts aren't supported in fontconfig and difficult to figure out an icon and a screenshot from unknown/empty "language" symbol.

Some of them are a minor language or one of dialects. We can't assign any language symbol for them as a primary script.  We may need something script-based maps perhaps.

hughsie (approved):

